PR #5, opened 2022-11-17
Add proxy spec file

emmawillis:
This is a first draft to add to as we work on the node proxy

### DIFF
--- a/proxies/proxy_spec.md
+++ b/proxies/proxy_spec.md
@@ -1,0 +1,116 @@
+
+# Test Harness Proxy Servers
+
+##General Guidelines
+
+#### Entity Types
+- In addition to returning a JSON of the object properties whenever an object is created, the proxy server needs to include the class of the object in the response. It should translate all SDK-specific class names to a recognized entityType when reporting the type returned by a command. 
+    - **entityType**: one of 'Variable', 'User', 'Feature', or 'Object'`
+
+#### Locations
+- Whenever a request results in an object being created, it should be stored in such a way that it can be referenced later with a location string
+- The location of the object should be indexed by the command that created the object, rather than an assumption about the type of the object. So an object returned by the `variable` command should be referenced by `variable/:id`
+
+#### Errors
+- All endpoints should be wrapped in a try/catch. If an exception is thrown, the response body should be a JSON object containing:
+    - **exception**: the exception message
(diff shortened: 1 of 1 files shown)


proxies/proxy_spec.md
```diff
@@ -0,0 +1,116 @@
...
+    name: string,
+    isAnonymous: boolean,
+    ...
+}
+```
+
+The server should respond with the same [Response Format](####Response) as below.
+
+### POST /:location 
+Call a function on a specific object, referenced by location (for example, `client/:clientId` or `command/variable/:variableId`).
```
ajwootto, line 65: the second example here seems to imply that there's a `/command/variable` endpoint, but with the current spec wouldn't this be more like `client/:clientId/variable` or something since you'd be called the `variable` command on the client's `location`?

Also is it a bit weird that the command is defined in the POST body here but then seems to map to a URL REST-style resource later on?

proxies/proxy_spec.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+## Server Endpoints
+
+### GET /spec
+The proxy should return a 200 status to indicate that the service has started. It should also return a JSON object in the response body, with the following properties:
+- **name**: name of the SDK being tested,
+- **version**: The version string of the SDK,
+- **capabilities**: An array of strings describing optional features that this SDK supports 
+The current capability options are `"edgeDb", "sse", "cloudBucketing"  `
```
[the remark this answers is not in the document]
emmawillis: oh oops that was supposed to be localBucketing, fixed!

proxies/proxy_spec.md
emmawillis: oh yeah that's kind of confusing, that was supposed to be an example of referencing the variable object rather than calling the variable function
so you'd send a request to `/client/clientId` with `command: variable` in the body
then if you wanted to call onUpdate on that variable after, you'd send a request to `/variable/variableId` with `command: onUpdate` in the body 

i was thinking the structure where we store the objects on the proxy would look like 
```
commands: {
     variable: {
         0: {....},
     },
}
```
where we store the return values of each command under the command name. So that's why i added 'command' to the location string, but that's definitely confusing so i think the location should just be `/variable/variableId`, not `/commands/variable/variableId`

emmawillis:
moved this to a notion!

PR #12, opened 2022-11-18
Dvc 5164 client api

emmawillis:


proxies/nodejs/app.ts
```diff
@@ -1,20 +1,53 @@
...
+  if (body.clientId === undefined) {
+    ctx.status = 400 
+    ctx.body = "Invalid request: Missing clientId"
+  } else if (!body.sdkKey) {
+    ctx.status = 400 
+    ctx.body = "Invalid request: Missing sdkKey"
+  } else {
+    const client = new DVCClient(ctx.request.body.sdkKey, ctx.request.body.options)
+    data.clients[ctx.request.body.clientId] = client
+    await client.onClientInitialized()
```
ajwootto, line 23: I dont think you want to wait for this in the client creation endpoint, because we need to be able to do tests that call methods on the client before its finished initializing. Waiting for init would be one of the `:location` requests

proxies/nodejs/app.ts
```diff
@@ -1,20 +1,53 @@
...
+}
+
+const handleClient = async (ctx) => {
+  const body = ctx.request.body
+  if (body.clientId === undefined) {
+    ctx.status = 400 
+    ctx.body = "Invalid request: Missing clientId"
+  } else if (!body.sdkKey) {
+    ctx.status = 400 
+    ctx.body = "Invalid request: Missing sdkKey"
```
ajwootto, line 19: do you want the api to validate this or do you want to just pass it verbatim to the SDK and see what happens? otherwise how do we test how the SDK handles not providing an SDK token?

proxies/nodejs/app.ts
emmawillis: good point, i'll take it out

emmawillis: also i screwed up the branch so i moved this to a new pr https://github.com/DevCycleHQ/test-harness/pull/13

proxies/nodejs/app.ts
```diff
@@ -1,20 +1,53 @@
...
+  } else {
+    const client = new DVCClient(ctx.request.body.sdkKey, ctx.request.body.options)
```
ajwootto, line 21: hmm, I think that the intention of the NodeJS SDK is that the end-user calls "initialize" rather than constructing the client directly like this. We should be calling whatever the method we expect the user to interact with here.

Also if possible we should probably prevent this style of creation of the client from being externally accessible, possibly by just exporting the type of DVCClient rather than the whole thing

emmawillis:
moved to https://github.com/DevCycleHQ/test-harness/pull/13

PR #13, opened 2022-11-18
Add client endpoint to proxy

emmawillis:


proxies/nodejs/app.ts
```diff
@@ -1,14 +1,37 @@
...
+    data.clients[ctx.request.body.clientId] = client
+    await client.onClientInitialized()
+
+    ctx.status = 200
+  }
+}
 
+async function start() {
     const app = new Koa()
+    app.use(KoaBody())
```
ajwootto, line 31: kinda confused at the difference between this and 
https://github.com/koajs/bodyparser

suthar26 (approved):
Looks good

suthar26:
@elliotCamblor just realized the location string is not being passed back, only the status is set to 200. unless if I am missing something

PR #14, opened 2022-11-18
add shared types file

elliotCamblor:


types/capabilities.ts
```diff
@@ -0,0 +1,6 @@
+export enum Capabilities {
```
ajwootto, line 1: shouldnt this be in some format that works cross language? like a JSON file or something?

elliotCamblor: yeah youre right lol.

jsalaber (approved):

PR #22, opened 2022-11-22
[DVC-5168] feat: add location call to proxy server 

suthar26:
- call methods and format return values
- assumptions: 
  - the call will only be made on `client` and `user` at the moment, 
  - callbacks not yet supported 
  - the POST `/location` call is being on a function that exists on the type
  - entityType just shows object, will be accurate after all object types are well defined 
  
  Tested on the following test data:
  
  ```js
      const testClientData = [
      {
        "clientId": 1,
        "sdkKey": "<SDKKEY>",
        "options": {},
      },
      {
        clientId: 2,
        sdkKey: "<SDKKEY>",
        options: {},
      }
    ]

    const testUserData = [
      {
        "user_id": "1",
        "name": "user1",
        isAnonymous: false,
      },
      {
        "user_id": "2",
        "name": "user2",
        isAnonymous: false,
      }
    ]
    testClientData.forEach(async (clientData) => {
      const client = await initialize(clientData.sdkKey, clientData.options).onClientInitialized()
      data.clients[clientData.clientId] = client
    }) 
    testUserData.forEach(async (userData) => {
      const user = <DVCUser>userData
      const userId = Object.keys(data.users).length
      data.users[userId] = user
    })

  ```
  using the following call:
  ```
  POST localhost:3000/client/2
  body : {
    command: variable
    params:[{"location": "user/1"},{"value": "variable-key"},{"value":"defaultValue"},{"callbackURL": "call back url string here"}]
  }
  ```

### DIFF
--- a/proxies/nodejs/app.ts
+++ b/proxies/nodejs/app.ts
@@ -4,6 +4,7 @@ import Router from "koa-router";
 import bodyParser from "koa-bodyparser";
 import { handleUser } from "./handlers/user";
 import { handleClient } from "./handlers/client";
+import { handleLocation } from "./handlers/location";
 
 type Data = {
   clients: { [key: string]: DVCClient };
@@ -21,23 +22,26 @@ async function start() {
   const app = new Koa();
   app.use(bodyParser());
 
-    const router = new Router()
+  const router = new Router();
 
-    router.get('/spec', (ctx) => {
-        ctx.status = 200
-        ctx.body = {
-            name: 'NodeJS',
-            version: '', // TODO add branch name or sdk version here
-            capabilities: ['EdgeDB', 'LocalBucketing'],
-        }
-    })
+  router.get("/spec", (ctx) => {
+    ctx.status = 200;
+    ctx.body = {
+      name: "NodeJS",
+      version: "", // TODO add branch name or sdk version here
+      capabilities: ["EdgeDB", "LocalBucketing"],
+    };
+  });
 
-    router.post('/client', (ctx: Koa.ParameterizedContext) => {
-        handleClient(ctx, data.clients)
-    })
-    router.post('/user', (ctx: Koa.ParameterizedContext) => {
-        handleUser(ctx, data.users)
-    })
+  router.post("/client", (ctx: Koa.ParameterizedContext) => {
+    handleClient(ctx, data.clients);
+  });
+  router.post("/user", (ctx: Koa.ParameterizedContext) => {
+    handleUser(ctx, data.users);
+  });
+  router.post("/:location*", (ctx: Koa.ParameterizedContext) => {
+    handleLocation(ctx, data);
+  });
 
   app.use(router.routes()).use(router.allowedMethods());
 
--- a/proxies/nodejs/app.ts
+++ b/proxies/nodejs/app.ts
@@ -1,25 +1,25 @@
-import { DVCClient, DVCUser } from '@devcycle/nodejs-server-sdk'
-import Koa from 'koa'
-import Router from 'koa-router'
-import bodyParser from 'koa-bodyparser'
-import { handleUser } from './handlers/user'
-import { handleClient } from './handlers/client'
+import { DVCClient, DVCUser } from "@devcycle/nodejs-server-sdk";
+import Koa from "koa";
+import Router from "koa-router";
+import bodyParser from "koa-bodyparser";
+import { handleUser } from "./handlers/user";
+import { handleClient } from "./handlers/client";
 
 type Data = {
-  clients: { [key: string]: DVCClient }
-  users: { [key: string]: DVCUser }
-  commandResults: { [key: string]: any }
-}
+  clients: { [key: string]: DVCClient };
+  users: { [key: string]: DVCUser };
+  commandResults: { [key: string]: any };
+};
 
 const data: Data = {
-    clients: {},
-    users: {},
-    commandResults: {}
-}
+  clients: {},
+  users: {},
+  commandResults: {},
+};
 
 async function start() {
-    const app = new Koa()
-    app.use(bodyParser())
+  const app = new Koa();
+  app.use(bodyParser());
 
     const router = new Router()
 
@@ -39,11 +39,11 @@ async function start() {
         handleUser(ctx, data.users)
     })
 
-    app.use(router.routes()).use(router.allowedMethods())
+  app.use(router.routes()).use(router.allowedMethods());
 
-    // Server!
-    console.log('Server started on port 3000')
-    app.listen(3000)
+  // Server!
+  console.log("Server started on port 3000");
+  app.listen(3000);
 }
 
-start()
+start();
--- a/proxies/nodejs/app.ts
+++ b/proxies/nodejs/app.ts
@@ -1,16 +1,11 @@
-import { DVCClient, DVCUser } from '@devcycle/nodejs-server-sdk'
 import Koa from 'koa'
 import Router from 'koa-router'
 import bodyParser from 'koa-bodyparser'
 import { handleUser } from './handlers/user'
 import { handleClient } from './handlers/client'
 import { handleLocation } from './handlers/location'
+import { Data } from './entityTypes'
 
-type Data = {
-  clients: { [key: string]: DVCClient }
-  users: { [key: string]: DVCUser }
-  commandResults: { [key: string]: any }
-}
 
 const data: Data = {
     clients: {},
--- a/proxies/nodejs/app.ts
+++ b/proxies/nodejs/app.ts
@@ -7,10 +7,10 @@ import { handleClient } from './handlers/client'
 import { handleLocation } from './handlers/location'
 
 type Data = {
-  clients: { [key: string]: DVCClient };
-  users: { [key: string]: DVCUser };
-  commandResults: { [key: string]: any };
-};
+  clients: { [key: string]: DVCClient }
+  users: { [key: string]: DVCUser }
+  commandResults: { [key: string]: any }
+}
 
 const data: Data = {
     clients: {},
--- a/proxies/nodejs/app.ts
+++ b/proxies/nodejs/app.ts
@@ -3,9 +3,8 @@ import Router from 'koa-router'
 import bodyParser from 'koa-bodyparser'
 import { handleUser } from './handlers/user'
 import { handleClient } from './handlers/client'
-import { handleLocation } from './handlers/location'
+import { validateLocationRequest } from './handlers/location'
 import { Data } from './entityTypes'
-
 
 const data: Data = {
     clients: {},
@@ -35,7 +34,7 @@ async function start() {
         handleUser(ctx, data.users)
     })
     router.post('/:location*', (ctx: Koa.ParameterizedContext) => {
-        handleLocation(ctx, data)
+        validateLocationRequest(ctx, data)
     })
 
     app.use(router.routes()).use(router.allowedMethods())
--- a/proxies/nodejs/app.ts
+++ b/proxies/nodejs/app.ts
@@ -1,10 +1,10 @@
-import { DVCClient, DVCUser } from "@devcycle/nodejs-server-sdk";
-import Koa from "koa";
-import Router from "koa-router";
-import bodyParser from "koa-bodyparser";
-import { handleUser } from "./handlers/user";
-import { handleClient } from "./handlers/client";
-import { handleLocation } from "./handlers/location";
+import { DVCClient, DVCUser } from '@devcycle/nodejs-server-sdk'
+import Koa from 'koa'
+import Router from 'koa-router'
+import bodyParser from 'koa-bodyparser'
+import { handleUser } from './handlers/user'
+import { handleClient } from './handlers/client'
+import { handleLocation } from './handlers/location'
 
 type Data = {
   clients: { [key: string]: DVCClient };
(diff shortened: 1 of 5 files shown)


proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,95 @@
...
+        data
+      );
+
+      // TODO: handle async commands
+      // TODO: handle callback before invoking command
+      const resultData = entity[command](...params);
+
+      ctx.status = 200;
+      ctx.body = {
+        entityType: typeof resultData, // assuming this is the type of the entity for returned data
```
nsalamad, line 49: `entityType` has to be one `Variable`, `User`, `Feature`, or `Object`. The reason being is that the class names vary across the SDKs  (DVCVariable Vs. Variable) so we need to have them agree on the same entity types. This way the test harness won't have to worry about the SDKs individually

suthar26: I was waiting on the `EntityTypes` are added to my code to ensure that that can be used to find the correct corresponding type for the saved objects. 

For the the function outputs, we may need to find a way to map the types to the command types to ensure we are not getting `any` response resulting in `Object` type

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,95 @@
...
+        data: resultData,
+        logs: [], // TODO add logs here
+      };
+    } catch (error) {
+      if (body.isAsync) {
+        ctx.status = 500;
```
nsalamad, line 55: The status here should still be 200 cause from the a testing perspective an error is a success cause the SDK did what it was supposed to do

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,95 @@
...
+        ctx.body = {
+          asyncError: error.message,
+          errorCode: error.code,
+        };
+      } else {
+        ctx.status = 500;
```
nsalamad, line 61: same here

proxies/nodejs/handlers/location.ts
suthar26: that is why i had it as 200, i changed it this morning :trollface:

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,95 @@
...
+        data
+      );
+
+      // TODO: handle async commands
+      // TODO: handle callback before invoking command
+      const resultData = entity[command](...params);
+
+      ctx.status = 200;
+      ctx.body = {
+        entityType: resultData.constructor.name, // assuming this is the type of the entity for returned data
```
nsalamad, line 49: This will give you the class name i.e. `DVCVariable` so you still need to convert it to one of the defined entity types. If the class name is `DVCVariable` then `entityType` should be `Variable`

.github/workflows/config.yml
```diff
@@ -1,6 +1,8 @@
-name: Node.js CI
+name: Run Harness Tests
```
jsalaber, line 1: did you rebase on main? looks like this isn't supposed to be here

jsalaber (changes requested):

.github/workflows/config.yml
suthar26: maybe i did it before that got merged in

suthar26: fixed

proxies/nodejs/handlers/location.ts
suthar26: fixed

proxies/nodejs/app.ts
```diff
@@ -1,49 +1,53 @@
-import { DVCClient, DVCUser } from '@devcycle/nodejs-server-sdk'
-import Koa from 'koa'
-import Router from 'koa-router'
-import bodyParser from 'koa-bodyparser'
-import { handleUser } from './handlers/user'
-import { handleClient } from './handlers/client'
+import { DVCClient, DVCUser } from "@devcycle/nodejs-server-sdk";
```
elliotCamblor, line 1: whats up with the linting in this PR? It looks like it should be single quotes and no semi-colons from what I can tell from the closed linting PR

suthar26:
added the saving of `commandResult` in the `data` global object.

New Test method:

make these calls in this order
- `POST /client` 
  - ```
    clientId:1
    sdkKey:<SDKKEY>
     ```
  - check for `Location` header to be `client/1`
- `POST /user` 
  - ```
    user_id:123
    name:testUser
     ```
  - check for `Location` header to be `user/0`
- `POST /client/1 ` `:location` call
  - ```
    command:variable
    params:[{"location": "user/0"},{"value": "variable key"},{ "value" : "default user"},{"callbackURL": "call back url string here"}]
    isAsync:true     
    ```
  - check for `Location` header to be `command/variable/0`
- the data object at the end of these calls is
```js
{
  clients: {
    '1': DVCClient {
      initialized: true,
      environmentKey: '<SDKKEY>',
      logger: [Object],
      onInitialized: [Promise],
      configHelper: [EnvironmentConfigManager],
      eventQueue: [EventQueue]
    }
  },
  users: { '0': { user_id: '123', name: 'testUser' } },
  commandResults: { variable: { '0': [DVCVariable] } }
}
```

proxies/nodejs/handlers/client.ts
```diff
@@ -8,23 +8,23 @@ type ClientRequestBody = {
 }
 
 export const handleClient = async (ctx: Koa.ParameterizedContext, clients: { [key: string]: DVCClient }) => {
-  const body = <ClientRequestBody>ctx.request.body
-  if (body.clientId === undefined) {
-    ctx.status = 400
-    ctx.body = {
-      error: "Invalid request: missing clientId"
+    const body = <ClientRequestBody>ctx.request.body
+    if (body.clientId === undefined) {
```
jsalaber, line 12: you can probably just return in this if statement after setting the error on the `ctx` obj so you can take the try catch out of the else statement

proxies/nodejs/app.ts
```diff
@@ -4,17 +4,18 @@ import Router from 'koa-router'
...
 import { handleUser } from './handlers/user'
 import { handleClient } from './handlers/client'
+import { handleLocation } from './handlers/location'
 
 type Data = {
-  clients: { [key: string]: DVCClient }
-  users: { [key: string]: DVCUser }
-  commandResults: { [key: string]: any }
-}
+  clients: { [key: string]: DVCClient };
```
jsalaber, line 10: a bunch of objects still have semicolons after them, make sure to remove them all

proxies/nodejs/app.ts
```diff
@@ -4,17 +4,18 @@ import Router from 'koa-router'
...
+  users: { [key: string]: DVCUser };
```
jsalaber, line 11: here too

proxies/nodejs/app.ts
```diff
@@ -4,17 +4,18 @@ import Router from 'koa-router'
...
+  commandResults: { [key: string]: any };
```
jsalaber, line 12: here too

proxies/nodejs/app.ts
```diff
@@ -4,17 +4,18 @@ import Router from 'koa-router'
...
+};
```
jsalaber, line 13: here

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,140 @@
+import Koa from 'koa'
+import { DVCClient, DVCUser } from '@devcycle/nodejs-server-sdk'
+import { getEntityModelFromType, getEntityNameFromType } from '../entityTypes'
+
+//HTTP request comes in as string
+type LocationRequestBody = {
+  command: string;
```
jsalaber, line 7: also these types

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,140 @@
...
+  params: string; //[{value: string | number | boolean} | {location: string} | {callbackUrl: string}]
```
jsalaber, line 8: and here

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,140 @@
+import Koa from 'koa'
+import { DVCClient, DVCUser } from '@devcycle/nodejs-server-sdk'
+import { getEntityModelFromType, getEntityNameFromType } from '../entityTypes'
+
+//HTTP request comes in as string
+type LocationRequestBody = {
+  command: string;
+  params: string; //[{value: string | number | boolean} | {location: string} | {callbackUrl: string}] 
+  isAsync: boolean;
```
jsalaber, line 9: and here

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,140 @@
...
+};
```
jsalaber, line 10: you get the idea

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,140 @@
...
+        const id = urlParts[urlParts.length - 1]
+        let entity
+        if (entityType === 'user') {
+            entity = data.users[id]
+        }
+        if (entityType === 'client') {
+            entity = data.clients[id]
+        }
+        return entity
+    } else if ( urlParts.length === 4) {
```
jsalaber, line 111: formatting on the space before `urlParts`

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,140 @@
...
+    data: {
+    clients: { [key: string]: DVCClient };
+    users: { [key: string]: DVCUser };
+    commandResults: { [key: string]: any };
+  }
+) => {
+    const body = ctx.request.body as LocationRequestBody
+    const entity = getEntityFromLocation(ctx.request.url, data)
+
+    if (entity === undefined) {
```
jsalaber, line 23: you can make these first two if statements guard statements as well, i.e. returning early so you can make the try catch on the root level of the function

proxies/nodejs/entityTypes.ts
```diff
@@ -1,6 +1,36 @@
...
-  variable = "Variable",
-  feature = "Feature",
-  object = "Object"
+  user = 'User',
+  variable = 'Variable',
+  feature = 'Feature',
+  object = 'Object',
+}
+
+export const getEntityModelFromType = (value: string): EntityTypes => {
```
jsalaber, line 8: if these functions don't get called individually without the other you can probably just combine them to return an `Entity` object that has both model and name

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,140 @@
...
+type LocationRequestBody = {
+  command: string;
+  params: string; //[{value: string | number | boolean} | {location: string} | {callbackUrl: string}] 
+  isAsync: boolean;
+};
+
+export const handleLocation = async (
+    ctx: Koa.ParameterizedContext,
+    data: {
+    clients: { [key: string]: DVCClient };
```
jsalaber, line 15: dunno if it's just github but formatting is weird here, should be indented

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,140 @@
...
+
+            const entityType = getEntityModelFromType(resultData.constructor.name)
+            const entityName = getEntityNameFromType(resultData.constructor.name)
+
+            const commandId = data.commandResults[entityName] !== undefined?
+                Object.keys(data.commandResults[entityName]).length :
+                0
+
+            if (data.commandResults[entityName] === undefined) {
+                data.commandResults[entityName] = {}
```
jsalaber, line 60: we can probably just define the data object with the commands it can store so we don't have to do this

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,140 @@
...
+    })
+    return parsedParams
+}
+
+const invokeCommand = async (entity, command, params, isAsync) => {
+    if (isAsync) {
+        const result = await entity[command](...params)
+        return result
+    }
+    {
```
jsalaber, line 137: why is this brace here? can just return it regardless, no need for an else

jsalaber (changes requested):

proxies/nodejs/handlers/client.ts
elliotCamblor: the try/catch in the else is intentional for testing error cases I believe

jsalaber: yeah i know i just meant you can flatten it by returning early in the if statements so it's easier to read the try catch

elliotCamblor: ah you mean so the whole function can just be wrapped?

jsalaber:
```
if (statement) {
    return
}

try {

} catch (e) {

}
```
more for code cleanliness, I think it makes it easier to read if there's less indentation :)

proxies/nodejs/handlers/location.ts
elliotCamblor: Thoughts on perhaps moving these checks to a middleware function? It could also attach the entity to the request body. It would really clean up the handler

proxies/nodejs/handlers/client.ts
elliotCamblor: I getcha

proxies/nodejs/handlers/location.ts
jsalaber: this middleware would probably be only specific to this endpoint, no?

elliotCamblor: yeah, its just a convenient way to separate the logic out

proxies/nodejs/handlers/location.ts
suthar26: that would only help if we are explicitly calling certain commands such as
 `user` and `variable` `onUpdate` `onInitialized`
 
 this model would allow for better future proofing?

proxies/nodejs/handlers/location.ts
suthar26: would you like me to add the middleware logic?

proxies/nodejs/entityTypes.ts
```diff
@@ -1,6 +1,34 @@
...
+
+export const getEntityFromType = (value: string): Entity => {
+    switch (value) {
+        case 'DVCUser':
+            return { type: EntityTypes.user, model: 'user' }
+        case 'DVCVariable':
+            return { type: EntityTypes.variable, model: 'variable' }
+        case 'DVCFeature':
+            return { type: EntityTypes.feature, model: 'feature' }
+        case 'DVCObject':
```
nsalamad, line 29: `DVCObject` doesn't exist in our sdk so you can just remove this

proxies/nodejs/handlers/location.ts
suthar26: added ✅

proxies/nodejs/handlers/client.ts
```diff
@@ -8,23 +8,23 @@ type ClientRequestBody = {
...
+        ctx.status = 400
+        ctx.body = {
+            error: 'Invalid request: missing clientId'
+        }
+    } else {
```
jsalaber, line 17: noticed you resolved the convo, did you decide against refactoring this out of the else clause?

proxies/nodejs/entityTypes.ts
```diff
@@ -1,6 +1,32 @@
...
+    model: string
+}
+
+export type Data = {
+    clients: { [key: string]: DVCClient }
+    users: { [key: string]: DVCUser }
+    commandResults: { [key: string]: any }
+}
+
+export const getEntityFromType = (value: string): Entity => {
```
suthar26, line 21: change to just having EntityType and add .lowercase for model uses

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,146 @@
...
+            ctx.body = {
+                errorCode: error.code,
+                exception: error.message,
+            }
+
+        }
+    }
+}
+
+const getEntityFromLocation = (location, data) => {
```
jsalaber, line 74: we don't have typescript linting yet, but could you add the types to the functions params?

proxies/nodejs/entityTypes.ts
suthar26: ✅

jsalaber (changes requested):

proxies/nodejs/handlers/client.ts
suthar26: oops, I only looked at `location.ts` and missed this

proxies/nodejs/handlers/location.ts
suthar26: added to all missing ones, hopefully they seem alright

proxies/nodejs/handlers/location.ts
```diff
@@ -0,0 +1,153 @@
...
+            parsedParams.push(element.value)
+        } else if (element.location !== undefined) {
+            parsedParams.push(getEntityFromLocation(element.location, data))
+        }
+    })
+    return parsedParams
+}
+
+const invokeCommand = async (
+    entity: DVCClient | DVCUser | DVCVariable | any,
```
jsalaber, line 119: once we have typescript linting we should remove any params that have `any` type, we should know explicitly what types a function can have

jsalaber (approved):

PR #23, opened 2022-11-23
chore: variable cloud tests

jsalaber:
# Summary

- Added helper methods to create client, user
- Added helper method to run tests for different variable types
- Added tests for cloud mode variable calls

### DIFF
--- a/harness/features/variable.cloud.test.ts
+++ b/harness/features/variable.cloud.test.ts
@@ -6,6 +6,8 @@ import nock from 'nock'
 
 jest.setTimeout(10000)
 
+const scope = getServerScope()
+
 describe('Variable Tests - Cloud', () => {
     forEachSDK((name) => {
         let url: string
@@ -14,17 +16,17 @@ describe('Variable Tests - Cloud', () => {
 
         beforeAll(async () => {
             url = getConnectionStringForProxy(name)
-            const res = await fetch(`${url}/spec`)
-            const response = await res.json()
+            // const res = await fetch(`${url}/spec`)
+            // const response = await res.json()
 
-            expect(response.name).toBeDefined()
-            expect(response.capabilities).toBeDefined()
+            // expect(response.name).toBeDefined()
+            // expect(response.capabilities).toBeDefined()
             
-            await createClient(url, clientId, 'SDK_KEY')
+            await createClient(url, clientId, 'server-939ecf39-866f-4682-8bfe-3639773b0fce')
         })
 
         describeIf(capabilities.includes(Capabilities.cloud))(name, () => {
-            it.only('will return default value if variable called with invalid user before initialized',  async () => {
+            it('will return default value if variable called with invalid user before initialized',  async () => {
                 const response = await createUser(url, { name: 'invalid user' })
                 const invalidUser = await response.json()
                 
@@ -43,7 +45,7 @@ describe('Variable Tests - Cloud', () => {
                 expect(variable.data.value).toBe('default_value')
             })
 
-            it('will throw error if variable called with invalid user after initialized',  async () => {
+            it.only('will throw error if variable called with invalid user after initialized',  async () => {
                 const response = await createUser(url, { name: 'invalid user' })
                 const invalidUser = await response.json()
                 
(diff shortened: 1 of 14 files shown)


harness/helpers.ts
```diff
@@ -22,4 +22,25 @@ export const forEachSDK = (tests) => {
...
+}
+
+export const createUser = async (url: string, user: object) => {
+    const response = await fetch(`${url}/user}`, {
+        method: 'POST',
+        body: JSON.stringify({
+            ...user
+        })
+    })
+    return response.text()
```
nsalamad, line 43: any reason for not doing `.json()` instead?

nsalamad: actually I think we should return the entire response object cause the tests will need to get the `Location` header

harness/helpers.ts
```diff
@@ -16,10 +16,36 @@ export const forEachSDK = (tests) => {
...
         SDKs = JSON.parse(process.env.SDKS_TO_TEST)
     } catch (e) {
-        console.warn('Unable to parse SDKS to test env var')
+        console.log('No specified SDKs to test, running all tests')
         SDKs = Object.values(Sdks)
     }
     describe.each(SDKs)('%s SDK tests', tests)
 }
 
+export const createClient = async (url: string, clientId: string, sdkKey: string, options?: object) => {
```
suthar26, line 25: should we make the `clientId` and `sdkKey` optional so those tests can fail?

harness/features/variable.cloud.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+describe('Variable Tests - Cloud', () => {
+    forEachSDK((name) => {
+        let url: string
+        const capabilities: string[] = SDKCapabilities[name]
+        const clientId: string = uuidv4()
+
+        beforeAll(async () => {
+            url = getConnectionStringForProxy(name)
+            await createClient(url, clientId, 'dvc_server_test_token')
```
suthar26, line 19: do you not need to add the option `{ enableCloudBucketing: true }`

jsalaber: yep, i'll commit it to the branch

harness/features/variable.cloud.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    callVariable, 
+    forEachVariableType,
+    variablesForTypes
+} from '../helpers'
+import { Capabilities, SDKCapabilities } from '../types'
+import { v4 as uuidv4 } from 'uuid'
+import { getServerScope } from '../mockServer'
+import nock from 'nock'
+
+jest.setTimeout(1000000)
```
nsalamad, line 17: isn't this too high?

jsalaber: yep, i think elliot changed it for testing and i forgot to change it back

harness/helpers.ts
```diff
@@ -16,10 +16,98 @@ export const forEachSDK = (tests) => {
...
+                { location: `${userLocation}` },
+                { value: key },
+                { value: defaultValue }
+            ]
+        })
+    })
+}
+
+export const wait = (ms: number) => {
+    return new Promise((resolve) => {
```
nsalamad, line 106: You can do `Promise<void>((resolve) => ....` and then you'll be able to call resolve without any args

harness/features/variable.cloud.test.ts
```diff
@@ -0,0 +1,260 @@
...
+import nock from 'nock'
+
+jest.setTimeout(10000)
+
+const scope = getServerScope()
+
+describe('Variable Tests - Cloud', () => {
+    forEachSDK((name) => {
+        let url: string
+        const mockServerUrl
```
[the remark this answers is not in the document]
jsalaber: this should be the base url of the mock server, not the proxy server, so it'll redirect to `${mockServerUrl}/v1/variables/key`

harness/features/variable.cloud.test.ts
```diff
@@ -0,0 +1,260 @@
...
+            it('will throw error variable called with invalid user',  async () => {
+                const response = await createUser(url, { name: 'invalid user' })
+                const invalidUser = await response.json()
+
+                expect(invalidUser.entityType).toBe('User')
+                expect(invalidUser.data.user_id).toBeUndefined()
+                expect(invalidUser.data.name).toBe('invalid user')
+
+                const userId = response.headers.get('location')
+                expect(userId.includes('user/')).toBeTruthy()
```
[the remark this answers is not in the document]
jsalaber: if we end up having proxy server tests then it'd be less useful, but as of right now i think it's fine to leave them

harness/features/variable.cloud.test.ts
```diff
@@ -0,0 +1,260 @@
...
+                    })
+                    .reply(200, {
+                        key: 'var_key',
+                        value: 'value',
+                        defaultValue: 'default_value',
+                        isDefaulted: false
+                    })
+                const variableResponse = await callVariable(newClientId, url, userId, 'var_key', 'default_value')
+                await variableResponse.json()
+                await wait(1000)
```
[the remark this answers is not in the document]
jsalaber: they were used to wait a bit before expecting `scope.isDone()` to be true, but I can remove it to see if it works without it

harness/features/variable.cloud.test.ts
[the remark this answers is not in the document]
elliotCamblor: yea I just thought the mock server was going to use the same format so that every mock is done by client Id as well. It might not actually be necessary but I think thats what the others are doing in their tests

elliotCamblor (approved):

PR #26, opened 2022-11-23
[DVC - 5171] handle callbacks and test with dummy server

suthar26:
since there are no callback parameters for the nodejs sdk, `.then()` seems to be the best way to initial callback to the URL

not using the following test data but making these calls instead to setup for the `:location` call

- `POST /client` 
  - ```
    clientId:1
    sdkKey:<SDKKEY>
     ```
  - check for `Location` header to be `client/1`
- `POST /user` 
  - ```
    user_id:123
    name:testUser
     ```
  - check for `Location` header to be `user/0`

test data:
```
  const testClientData = [
    {
      clientId: 1,
      sdkKey: "SDK_KEY",
      options: {},
    },
    {
      clientId: 2,
      sdkKey: "SDK_KEY",
      options: {},
    },
  ];

  const testUserData = [
    {
      user_id: "1",
      name: "user1",
      isAnonymous: false,
    },
    {
      user_id: "2",
      name: "user2",
      isAnonymous: false,
    },
  ];
  testClientData.forEach(async (clientData) => {
    const client = await initialize(
      clientData.sdkKey,
      clientData.options
    ).onClientInitialized();
    data.clients[clientData.clientId] = client;
  });
  testUserData.forEach(async (userData) => {
    const user = <DVCUser>userData;
    const userId = Object.keys(data.users).length;
    data.users[userId] = user;
  });

  router.post("/update", (ctx) => {
    console.log(ctx);
    ctx.status = 200;
    ctx.body = {
      data: "update",
    };
  });

```

test call - check comment for the updated test call: 
```
POST localhost:3000/client/2
body : {
  command: variable
  params:[{"location": "user/1"},{"value": "variable-key"},{"value":"defaultValue"},{"callbackURL": "localhost:8000/update"}]
}
```

setup a test express server on `PORT 8000` to capture the calback


suthar26:
Test calls:

- POST /client/1 :location call
  - ```
    command:variable
    params:[{"location": "user/0"},{"value": "test-str-variable"},{"value":"checking for variable value"},{"callbackURL": "http://localhost:8000/update"}]
    isAsync:false    
    ```
  - check for the console log for 
  ```
    onUpdatecallback data  {
      entityType: 'Variable',
      data: {
        key: 'test-str-variable',
        isDefaulted: false,
        value: 'hi this variable is off',
        defaultValue: 'checking for variable value'
      }
    }
  ```
  this is what the dummy server received, I am just sending it back into the `res.json`
  
  making sure the non call back calls also still work
- POST /client/1 :location call
  - ```
    command:variable
    params:[{"location": "user/0"},{"value": "test-str-variable"},{"value":"checking for variable value"}]
    isAsync:false    
    ```
  - this will return  
  ```
  {
    "entityType": "Variable",
    "data": {
        "key": "test-str-variable",
        "isDefaulted": false,
        "value": "hi this variable is off",
        "defaultValue": "checking for variable value"
    },
    "logs": []
  }
  ```

proxies/nodejs/handlers/location.ts
```diff
@@ -24,37 +25,56 @@ export const handleLocation = async (
...
+                    })
+                    .then((resp: any) => console.log('onUpdatecallback data ', resp.data))
+                    .catch((e) => console.error(e))
+            }
+            invokeCommand(entity, command, params, body.isAsync).then((data) => {
+                onUpdateCallback(data)
+            })
+            ctx.status = 200
+            ctx.body = {
+                entityType: 'pending', // TODO: add pending entity type
```
jsalaber, line 46: if these don't make sense to have in the return statement we don't need to include it as pending

proxies/nodejs/handlers/location.ts
```diff
@@ -24,37 +25,56 @@ export const handleLocation = async (
             JSON.parse(body.params),
             data
         )
+        const lastParam = params[params.length - 1]
+
+        if (lastParam instanceof URL) {
```
jsalaber, line 30: is there a reason to convert this to a URL object? seems unnecessary if you're just gonna use the url with `callbackURL.href`

jsalaber (changes requested):

proxies/nodejs/handlers/location.ts
suthar26: converting to URL since all the other values will be `< string | boolean | number >` so gives a better representation of the check

proxies/nodejs/handlers/location.ts
suthar26: any idea on what should return object be?

proxies/nodejs/handlers/location.ts
```diff
@@ -9,7 +10,7 @@ type LocationRequestBody = {
     isAsync: boolean
 }
 
-type parsedParams = (string | boolean | number | object)[];
+type parsedParams = (string | boolean | number | object | URL)[];
```
nsalamad, line 13: nitpick: type names should be in PascalCase

proxies/nodejs/handlers/location.ts
jsalaber: could just send the logs instead of returning an entitytype and data

suthar26: Logger only exists on the `DVCClient` object, that would limit the usage.

sending the logs does not work as expected, 
```
            console.log('logger', entity.logger.info())
            console.log('logger', entity.logger.info('test'))
            console.log('logger', entity.logger.info)
```
outputs 
```
logger undefined
logger undefined
logger [Function: noOpLog]
```

couldn't find a function call we can make to get the logs..

jsalaber: well in either case i don't see a scenario where an object has a param that is a callback but also returns an object

suthar26:
Update: 
- For the current scope(nodejs-sdk) we only have the use for onClientInitialized  callback function,
- it makes sense for the onClientInitializedCallbackURL to be passed in when creating the client. otherwise, the client get initialized too quickly to attach a onClientInitialized through the /:location endpoint and would be very hard to test.
- closing this PR, keep the work in the archive for when we are working on sdks that have other non-race condition callbacks. (DVCVariable.onUpdate)

suthar26:
repoening because of changed implementation which matches the intended purpose

elliotCamblor (approved):

jsalaber (approved):

PR #29, opened 2022-11-30
Dvc 5176 initialize tests

nsalamad:


suthar26 (approved):
Tests look good

harness/features/initialize.local.test.ts
```diff
@@ -0,0 +1,182 @@
...
+                const responesBody = await response.json()
+
+                expect(responesBody.exception).toEqual(
+                    'Invalid environment key provided. Please call initialize with a valid server environment key'
+                )
+            })
+
+            it('initializes correctly on valid data', async () => {
+                const clientId = uuidv4()
+                const response = await createClient(url, clientId, sdkKey)
```
elliotCamblor, line 50: since theres no baseURLOverride in here, the proxy server will try and make real requests to our prod api instead of the mock server

nsalamad: good catch

harness/features/initialize.local.test.ts
```diff
@@ -0,0 +1,179 @@
...
+                const clientId = uuidv4()
+                const response = await createClient(url, clientId)
+                const responesBody = await response.json()
+
+                expect(responesBody.exception).toEqual(
+                    'Missing environment key! Call initialize with a valid environment key'
+                )
+            })
+
+            it('should error when SDK key is invalid', async () => {
```
elliotCamblor, line 38: I think this test and the above should still mock out the config request, and then assert that a config request wasn't made. Just to test that the SDK token validation is being done in the SDK rather than server side

harness/features/initialize.local.test.ts
```diff
@@ -0,0 +1,179 @@
...
+                const clientId = uuidv4()
+                const response = await createClient(url, clientId, 'invalid key')
+                const responesBody = await response.json()
+
+                expect(responesBody.exception).toEqual(
+                    'Invalid environment key provided. Please call initialize with a valid server environment key'
+                )
+            })
+
+            it('initializes correctly on valid data', async () => {
```
elliotCamblor, line 48: Probably worth asserting that a config request was made for this case

harness/features/initialize.local.test.ts
```diff
@@ -0,0 +1,179 @@
...
+                const callbackSubdirectory = `/client/${clientId}/onClientInitialized`
+                scope
+                    .post(callbackSubdirectory, { message: `onClientInitialized was invoked on /client/${clientId}` })
+                    .matchHeader('Content-Type', 'application/json')
+                    .reply(204)
+                
+                await createClient(url, clientId, sdkKey)
+                await callOnClientInitialized(clientId, url, `${harnessBaseURL}${callbackSubdirectory}`)
+
+                await new Promise<void>((resolve) => {
```
elliotCamblor, line 71: I believe Jason added a `wait` helper function so you wont need these Promise timeouts everywhere

harness/features/initialize.local.test.ts
```diff
@@ -0,0 +1,176 @@
...
+                    .reply(204)
+
+                await createClient(url, clientId, sdkKey, { baseURLOverride: `${harnessBaseURL}/client/${clientId}` })
+                await callOnClientInitialized(clientId, url, `${harnessBaseURL}${callbackSubdirectory}`)
+                await wait(500)
+
+                expect(scope.isDone()).toBeTruthy()
+            })
+
+            it('fetches config again after 10 seconds', async () => {
```
elliotCamblor, line 119: debating if we even need this case since youre also testing it below, and 10 seconds is a long time to have to wait in a test suite. Thoughts?

harness/features/initialize.local.test.ts
```diff
@@ -0,0 +1,153 @@
...
+        })
+
+        describeIf(capabilities.includes(Capabilities.local))(name, () => {
+            it('should error when SDK key is missing', async () => {
+                const clientId = uuidv4()
+                scope
+                    .get(`/client/${clientId}/config/v1/server/${sdkKey}.json`)
+                    .reply(200, {})
+
+                const response = await createClient(url, clientId)
```
jsalaber, line 41: you can likely just create one client for this test block if you don't specifically need it to be a different client, jest runs all tests in a test file sequentially

harness/features/initialize.local.test.ts
```diff
@@ -0,0 +1,153 @@
...
+                const { exception } = await response.json()
+                await wait(500)
```
jsalaber, line 43: I ended up not needing to wait, you might be able to just remove the waits

jsalaber (changes requested):

harness/features/initialize.local.test.ts
elliotCamblor: though since these are testing the initialize call I think it makes sense to create a separate client for each

harness/features/initialize.local.test.ts
nsalamad: For the first 2 tests the wait needs to be there so that the sdk has a chance fetch the config in case the error checks fail. Otherwise the test would report a false positive. The rest of the tests fail if I don't wait.

I think the reason you're tests pass is because in [handleLocation](https://github.com/DevCycleHQ/test-harness/blob/main/proxies/nodejs/handlers/location.ts#L57) `invokeCommand` would return a `Promise<DVCVariableInterface>` and we `await` it, so the proxy does not respond until after it fetches the variable from the mock endpoint.

I think the logic in `handleLocation` is incorrect because right now it awaits all skd methods regardless of weather you send the `isAsync` param

jsalaber: for this test do you need to wait for a config to come back if the sdk key is supposed to error it immediately?

jsalaber (changes requested):

harness/features/initialize.local.test.ts
nsalamad: hmm I guess not. If the skd fails to throw then the error message won't be there so that's a for sure way to check. I'll remove the wait and also the mock config endpoint

elliotCamblor (approved):

jsalaber (approved):

PR #31, opened 2022-11-30
fix up the middleware so it runs in the middle

elliotCamblor:


suthar26 (approved):

suthar26 (approved):

proxies/nodejs/app.ts
```diff
@@ -3,10 +3,10 @@ import Router from 'koa-router'
 import bodyParser from 'koa-bodyparser'
 import { handleUser } from './handlers/user'
 import { handleClient } from './handlers/client'
-import { validateLocationRequest } from './handlers/location'
+import { handleLocation, validateLocationReqMiddleware } from './handlers/location'
 import { Data } from './entityTypes'
 
-const data: Data = {
+export const dataStore: Data = {
```
nsalamad, line 9: could you rename the type to `DataStore` as well?

elliotCamblor: that seems reasonable

PR #33, opened 2022-12-01
Add variable local tests

emmawillis:


harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+                it('should throw exception if user is invalid',  async () => {    
+                    const variableResponse =
+                        await callVariable( clientId, url, invalidUser, 'string-var', 'string default')
+                    const variable = await variableResponse.json()
+
+                    expect(variable.exception).toBe('Must have a user_id set on the user')
+                })
+
+                it('should return default value if default type doesn\'t match variable type',  async () => {
```
jsalaber, line 119: i have made a helper method to be able to call a suite of tests with the different types of variables, so you can put the tests needing to test variable values / default values in one describe and use the helper method i made, `forEachVariableType`

jsalaber (changes requested):
Do you think aggregate events should be a part of these test suites as well?

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,193 @@
...
+    forEachSDK((name) => {
+        let url: string
+        const capabilities: string[] = SDKCapabilities[name]
+        const clientId: string = uuidv4()
+        const baseURL = `http://host.docker.internal:${global.__MOCK_SERVER_PORT__}/client/${clientId}`
+        let variationOnUser: string, noVariationUser: string, invalidUser: string
+
+        let resolveIsInitialized: (value: unknown) => void
+
+        const isInitialized = new Promise(function(resolve,){
```
elliotCamblor, line 31: this is jank but i guess it works lol

elliotCamblor: I think you have an extra comma in there though

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,193 @@
...
+            setTimeout(() => {
+                resolve({})
+            }, 1000)
+        })
+        expect(mockEvents).toHaveBeenCalledWith({
+            batch: [{ 
+                user: expect.objectContaining({
+                    platform: 'NodeJS',
+                    sdkType: 'server',
+                    // user_id: userId
```
elliotCamblor, line 162: how come this is commented out?

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,286 @@
...
+import { Capabilities, SDKCapabilities } from '../types'
+import { v4 as uuidv4 } from 'uuid'
+import { getServerScope } from '../mockServer'
+import { config } from '../mockData/index'
+
+jest.setTimeout(10000)
+
+const scope = getServerScope()
+
+const expectedVariablesByType = {
```
elliotCamblor, line 20: dont forget to add the type field to these

harness/features/variable.local.test.ts
emmawillis: added both!

harness/helpers.ts
```diff
@@ -118,17 +119,36 @@ export const wait = (ms: number) => {
...
-    return await fetch(`${url}/client/${clientId}`, {
+export const callOnClientInitialized = async (
+    clientId: string, 
+    proxyUrl: string, 
+    mockServerUrl: string,
+    scope: Scope
+) => {
+    let resolveIsInitialized: (value: unknown) => void
+
+    const isInitialized = new Promise(function(resolve){
```
ajwootto, line 130: might want to add a timeout to this so it doesn't hang indefinitely if never called

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,325 @@
...
+                        await callVariable( clientId, url, invalidUser.location, 'string-var', 'string default')
+                    const variable = await variableResponse.json()
+                    expectDefaultValue(variable, 'string default')
+                })
+            })
+
+            describe('initialized client', () => {
+                beforeAll(async () => {
+                    // mock endpoints
+                    scope
```
ajwootto, line 113: seems kind of weird to provide this request mock after the client has already been created earlier, what if it tries to hit this endpoint before this is set up?

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,325 @@
...
+                    
+                        const variableResponse = await callVariable(
+                            clientId, 
+                            url, 
+                            variationOnUser.location, 
+                            key, 
+                            defaultValue
+                        )
+                        const variable = await variableResponse.json()
+                        expect(variable.entityType).toBe('Variable')
```
ajwootto, line 147: might be easier to read these assertions as just a single object matcher like 
```
expect(variable).toEqual({
  isDefaulted: false
  ...
})
```

this would also let you be explicit about allowing / not allowing additional properties

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,325 @@
...
+                        expect(JSON.stringify(variable.data.value)).toBe(JSON.stringify(variationOn))
+                        expect(variable.data.isDefaulted).toBe(false)
+                        expect(JSON.stringify(variable.data.defaultValue)).toBe(JSON.stringify(defaultValue))
```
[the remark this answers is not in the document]
emmawillis: for the json variable tests the objects don't match unless they're stringified 
without it the expect fails with a warning saying 'objects serialize to the same string'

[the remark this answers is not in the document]
ajwootto: thats probably because you're using `toBe` instead of `toEqual`. `toBe` checks for referential equality

harness/features/variable.local.test.ts
ajwootto: you can add an `expect.matchObject` on that field or something

harness/features/variable.local.test.ts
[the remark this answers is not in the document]
emmawillis: oh oops gotcha

harness/features/variable.local.test.ts
emmawillis: yeah i just got the toEqual working and then it worked!

harness/features/variable.local.test.ts
emmawillis: true, i moved it to the beforeall

harness/features/initialize.local.test.ts
```diff
@@ -77,9 +77,10 @@ describe('Initialize Tests - Local', () => {
                     .matchHeader('Content-Type', 'application/json')
                     .reply(204)
 
-                await createClient(url, clientId, sdkKey, { baseURLOverride: `${mockServerUrl}/client/${clientId}` })
-                await callOnClientInitialized(clientId, url, `${mockServerUrl}${callbackSubdirectory}`)
-                await wait(500)
+                await createClient(url, clientId, sdkKey, { baseURLOverride: `${mockServerUrl}/client/${clientId}` })                
+                await callOnClientInitialized(clientId, url, mockServerUrl, scope)
+
+                expect(scope.isDone()).toBeTruthy()
```
ajwootto, line 83: you can get rid of these now

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,356 @@
...
+import { v4 as uuidv4 } from 'uuid'
+import { getServerScope } from '../nock'
+import { config } from '../mockData/index'
+import nock from 'nock'
+
+jest.setTimeout(10000)
+
+const scope = getServerScope()
+
+const expectedVariablesByType = {
```
ajwootto, line 22: maybe use the `BucketedConfig['variables']` type here like I added in the mock data

ajwootto: oh nvm I see this is not that type

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,356 @@
...
+    // these should match the config in mockData
```
ajwootto, line 23: does this mean like "the tests should prove that these match the config", or is it a note to us that the data defined here needs to match whats in the config? If so can you just use what's in the config or derive from it somehow?

harness/helpers.ts
```diff
@@ -135,17 +136,56 @@ export const wait = (ms: number) => {
...
+    const timeoutPromise = new Promise((_, reject) => {
+        setTimeout(() => {
+            reject(new Error(timeoutMessage))
+        }, timeout)
+    })
+
+    await Promise.race([
+        new Promise((resolve) => {
+            const callback =  (req, inter) => {
+                if (inter.uri === callbackURL) {
```
ajwootto, line 182: it might be better if this could take the actual interceptor you're waiting rather than just the URL. There may be times where youre using the same URL in two different interceptors but there are other conditions its matching on (headers, body contents etc.)

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,356 @@
...
+            value: ValueTypes,
+            isDefaulted: boolean,
+            defaultValue: ValueTypes
+        }
+    }
+    
+    const expectDefaultValue = (variable: VariableResponse, defaultValue: ValueTypes) => {
+        expect(variable).toEqual(expect.objectContaining({
+            entityType: 'Variable',
+            data: expect.objectContaining({
```
ajwootto, line 349: can this just be an exact match? currently this allows additional fields to be present on the variable, and we probably want to test the exact set of fields we're expecting

emmawillis: fixed!

harness/helpers.ts
emmawillis: sure, fixed!

harness/helpers.ts
```diff
@@ -135,17 +136,64 @@ export const wait = (ms: number) => {
     })
 }
 
-export const callOnClientInitialized = async (clientId: string, url: string, callbackURL: string) => {
-    return await fetch(`${url}/client/${clientId}`, {
+export const callOnClientInitialized = async (
+    clientId: string, 
+    proxyUrl: string,
```
ajwootto, line 141: you could also obtain this one just by knowing the SDK name right? using `getConnectionStringForProxy` ?

harness/helpers.ts
```diff
@@ -234,6 +220,11 @@ export class TestClient {
     }
 
     async callOnClientInitialized() {
+        const scope = getServerScope()
```
ajwootto, line 223: you should just leave this method as is, I deliberately didn't use the callback registration approach here because its already a promise-returning method so we can just send `isAsync: true`

harness/features/variable.local.test.ts
```diff
@@ -0,0 +1,328 @@
...
+        describe('uninitialized client', () => {
+            const testClient = new TestClient(name)
+
+            beforeAll(async () => {
+                await testClient.createClient()
+                const configRequestUrl = `/${testClient.clientLocation}/config/v1/server/${testClient.sdkKey}.json`
+                const interceptor = scope
+                    .get(configRequestUrl)
+                
+                interceptor.delay(2000).reply(404)
```
ajwootto, line 250: you don't need the delay here if you're intentionally making it 404 on the config request. I think you're conflating two different versions of the client being "uninitialized"
1. the client hasn't finished fetching the config yet
2. the client has finished trying to fetch the config and that fetch failed

I think what you're setting up here is case #2, but you're just delaying it by 2 seconds.

A delay would be useful for case #1, where you want to delay the config request from responding until after your test cases have run (meaning you'd have this delay but then not waitForRequest below)

emmawillis: oh yeah i was going for case 1, but i had to have it reply with something so i figured it was better to have the 404 as a backup. but then i added the waitForRequest because it was sometimes receiving the request in time (which broke the tests), but at that point the delay is pointless - fixed!!

ajwootto (approved):

jsalaber (approved):

PR #34, opened 2022-12-01
chore: add track tests

suthar26:


harness/features/track.local.test.ts
```diff
@@ -0,0 +1,378 @@
...
+const scope = getServerScope()
+
+describe('Track Tests - Local', () => {
+    const mockEvents = jest.fn()
+    const validUserId = 'user1'
+    forEachSDK((name) => {
+        let url: string
+        const capabilities: string[] = SDKCapabilities[name]
+        const clientId: string = uuidv4()
+        const mockServerUrl = `http://host.docker.internal:${global.__MOCK_SERVER_PORT__}`
```
nsalamad, line 19: this should be `http://${process.env.DOCKER_HOST_IP ?? 'host.docker.internal'}:${global.__MOCK_SERVER_PORT__}` :P

suthar26: ofcourse

harness/features/track.cloud.test.ts
```diff
@@ -0,0 +1,156 @@
...
+                        mockEvents(body)
+                        return [201, { success: true }]
+                    })
+
+                await wait(1000)
+                await trackResponse.json()
+
+                expect(mockEvents).toHaveBeenCalledTimes(1)
+                const track = mockEvents.mock.calls[0][0]
+                expect(track).toBeDefined()
```
ajwootto, line 69: this is a bit of a weird way to access this info since the way nock works should make it impossible for this to have been called more than once anyway. Might want to look at how Emma is doing it in her PR (where she just assigns the body to a variable directly)

harness/features/track.cloud.test.ts
```diff
@@ -0,0 +1,156 @@
...
+                const value = 1
+
+                const response = await createUser(url, { user_id: validUserId })
+                await response.json()
+                const userId = response.headers.get('location')
+
+                const trackResponse = await callTrack(clientId, url, userId,
+                    { type: eventType, target: variableId, value })
+
+                scope
```
ajwootto, line 96: shouldn't these be set up before you call track?

harness/features/track.cloud.test.ts
```diff
@@ -0,0 +1,156 @@
...
+                    .post(`/client/${clientId}/v1/track`)
+                    .matchHeader('Content-Type', 'application/json')
+                    .reply((uri, body) => {
+                        mockEvents(body)
+                        return [201, { success: true }]
+                    })
+
+                await wait(1000)
```
ajwootto, line 109: this should probably just wait for the appropriate requests to be called (see the discussion on slack about that)

harness/features/track.cloud.test.ts
```diff
@@ -0,0 +1,156 @@
...
+                expect(track.events[0].target).toBe(variableId)
+                expect(track.events[0].value).toBe(value)
+
+                expect(scope.isDone()).toBeTruthy()
+            })
+
+        })
+    })
+
+    const callTrack = async (clientId: string, url: string, userLocation: string, event: any) => {
```
ajwootto, line 133: should this go in `helpers` with the rest of the request methods?

ajwootto, line 133: doesnt matter much here but you should always use `unknown` instead of `any` to indicate "I dont know / care about the type". Using `unknown` prevents you from doing things to the value while not knowing what type it is, where `any` basically disables type checking entirely for that variable

harness/features/track.cloud.test.ts
```diff
@@ -0,0 +1,156 @@
...
+                command: 'track',
+                params: [
+                    { location: `${userLocation}` },
+                    { value: event }
+                ],
+            })
+        })
+    }
+
+    const wait = (ms: number) => {
```
ajwootto, line 149: is there already a helper function for this in `helpers`?

harness/features/track.local.test.ts
```diff
@@ -0,0 +1,369 @@
...
+        afterEach(() => {
+            jest.clearAllMocks()
+        })
+
+        // only for debugging
+        beforeEach(() => {
+            console.log('beforeEach', scope.activeMocks())
+        })
+
+        describeIf(capabilities.includes(Capabilities.cloud))(name, () => {
```
ajwootto, line 51: checking the wrong capability here

harness/features/track.local.test.ts
```diff
@@ -0,0 +1,369 @@
...
+            describe('Failing tests', () => {
```
ajwootto, line 52: is this just to keep track of tests that are currently known to be failing? Could use `it.failing` for that

harness/features/track.local.test.ts
```diff
@@ -0,0 +1,369 @@
...
+                it('should not send an event if the event type not set', async () => {
+
+                    const response = await createUser(url, { user_id: validUserId })
+                    await response.json()
+                    const userId = response.headers.get('location')
+
+                    const trackResponse = await callTrack(clientId, url, userId, {})
+
+                    scope.post(`/client/${clientId}/v1/events/batch`).reply((uri, body) => {
```
ajwootto, line 61: you could also try not setting up a scope at all here and using the `no match` event in Nock to assert that no unexpected requests were made:
https://github.com/nock/nock#global-no-match-event

this would let you use the `scope.isDone` check as well since you wouldn't have set up an expected request

harness/features/track.cloud.test.ts
suthar26: yes, but also works without. will move it for better readability and matching code patterns

harness/features/track.local.test.ts
suthar26: was nor the case, but fixing wording

harness/features/track.cloud.test.ts
```diff
@@ -0,0 +1,138 @@
+import { getConnectionStringForProxy, forEachSDK, describeIf, createClient, createUser, callTrack } from '../helpers'
+import { Capabilities, SDKCapabilities } from '../types'
+import { v4 as uuidv4 } from 'uuid'
+import { getServerScope } from '../nock'
+import nock from 'nock'
+
+jest.setTimeout(1000000)
```
ajwootto, line 7: thats a long timeout

suthar26: Copy pasta, reduced it to 10s and 15s

harness/types/sdks.ts
```diff
@@ -1,3 +1,5 @@
 export const Sdks = {
     nodejs: "NodeJS"
 }
+
+export const latestNodeJsSdkVersion = '1.4.23'
```
ajwootto, line 5: what's this for?

suthar26: old code no use, removed

harness/features/track.local.test.ts
```diff
@@ -0,0 +1,355 @@
...
+                it('should not send an event if the event type not set', async () => {
+                    let eventBody = {}
+
+                    const response = await createUser(url, { user_id: validUserId })
+                    await response.json()
+                    const userId = response.headers.get('location')
+
+                    const trackResponse = await callTrack(clientId, url, userId, {})
+
+                    nock.emitter.on('no match', (req) => {
```
ajwootto, line 65: if this is testing that no events are sent then this should never happen

harness/features/track.local.test.ts
```diff
@@ -0,0 +1,355 @@
...
+                        req.body = eventBody
+                    })
+
+                    await wait(eventFlushIntervalMS * 3) // wait for 2 event flush for safety
+
+                    const res = await trackResponse.json()
+                    // expect(res.exception).toBe('Missing parameter: type') // works for GH actions sometimes
```
ajwootto, line 72: this seems fishy

harness/features/track.local.test.ts
```diff
@@ -0,0 +1,355 @@
...
+                                ]
+                            },
+                        ],
+                    })
+                })
+
+                // this test is hard to test for and so the test is changed to ensure that there are 2 flushes
+                // that happen in the roughly specified flush time and also that there are no extra flush that
+                // take place. which is why this test will take longer to run
+                it('should retry with exponential backoff events API call to track 2 events', async () => {
```
ajwootto, line 245: change the description here since it isn't checking for exponential backoff anymore

harness/features/track.local.test.ts
suthar26: oh yes gets caught globally

ajwootto: yeah but it shouldn't hit an unmatched request in this test, since no flush request should be made (since the event is invalid)

harness/features/track.local.test.ts
suthar26: that is the problem i couldnt fix, ive had both of these be the response from the SDK. it SHOULD be `Missing parameter: type` but it is not thrown sometimes

ajwootto: one way this could happen is if the client was somehow not initialized by the time you call track, since it early returns in that case:
image

harness/features/track.local.test.ts
```diff
@@ -0,0 +1,268 @@
...
+                if (typeof body === 'object'
+                    && body.message.includes('onClientInitialized was invoked'))
+                    return [200]
+            })
+
+            await createClient(url, clientId, sdkKey, {
+                baseURLOverride: `${mockServerUrl}/client/${clientId}`,
+                eventFlushIntervalMS: eventFlushIntervalMS, logLevel: 'debug', configPollingIntervalMS: 1000 * 60
+            })
+            await callOnClientInitialized(clientId, url, `${mockServerUrl}/client/${clientId}`)
```
ajwootto, line 51: I think the problem here is that nothing is actually waiting for the client to finish initializing. This call is just "registering" the callback handler but nothing is actually waiting for that callback to be received. That's probably why you're seeing inconsistent behaviour when running the tests since there's basically a race condition.

The TestClient helper fixes this by changing the way it does onClientInitialized to just send a request with `isAsync` true. You could fix it here by either switching to use the TestClient helper or just update this existing `callOnClientInitialized` method to send "isAsync" true instead of trying to register a callback.

ajwootto: this should fix the check you had to comment out below

suthar26: makes sense, could just use waitForRequest with the `onClientInitialized interceptor` and fix this so that the client is initialized before any tests run.

ajwootto: you could, but its a bit of a roundabout approach since passing `isAsync` on the onClientInitialized request will do the same thing

harness/features/track.local.test.ts
suthar26: yeah i checked docker logs and saw the warn message

ajwootto (approved):

PR #38, opened 2022-12-06
feat: checkout sha of node js sdk if given from workflow

jsalaber:
# Summary

If a sha has been passed in from `workflow_dispatch`, then we write it to an `.env` file and use it to checkout a specific commit of the nodejs sdk and install that version after building with nx

### DIFF
--- a/harness/features/variable.cloud.test.ts
+++ b/harness/features/variable.cloud.test.ts
@@ -19,7 +19,7 @@ const scope = getServerScope()
 describe('Variable Tests - Cloud', () => {
     forEachSDK((name) => {
         let url: string
-        const capabilities: string[] = SDKCapabilities[name]
+        const capabilities: string[] = SDKCapabilities[name] || []
         const clientId: string = uuidv4()
         const sdkKey: string = `dvc_server_${clientId}`
         const mockServerUrl
--- a/harness/features/variable.cloud.test.ts
+++ b/harness/features/variable.cloud.test.ts
@@ -19,7 +19,7 @@ const scope = getServerScope()
 describe('Variable Tests - Cloud', () => {
     forEachSDK((name) => {
         let url: string
-        const capabilities: string[] = SDKCapabilities[name] || []
+        const capabilities: string[] = SDKCapabilities[name]
         const clientId: string = uuidv4()
         const sdkKey: string = `dvc_server_${clientId}`
         const mockServerUrl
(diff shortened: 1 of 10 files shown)


proxies/nodejs/install.sh
```diff
@@ -1,4 +1,12 @@
-if [ -z "$NODEJS_SDK_VERSION" ]; then
+if [ "GITHUB_SHA" ]; then
+    git clone https://github.com/DevCycleHQ/js-sdks.git
+    cd js-sdks
+    git checkout $GITHUB_SHA
+    yarn
+    NX_DAEMON=false yarn nx build nodejs --verbose
+    cd ..
+    yarn add file:js-sdks/dist/sdk/nodejs/
```
ajwootto, line 8: does this actually work without also adding all the dependent packages that are in this directory as well? Or is it installing those packages from NPM instead? Might want to make sure you're getting the local versions of all of them.

proxies/nodejs/install.sh
```diff
@@ -1,4 +1,12 @@
-if [ -z "$NODEJS_SDK_VERSION" ]; then
+if [ "GITHUB_SHA" ]; then
```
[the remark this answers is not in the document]
jsalaber: i think the idea was that if a SHA was specified, then it would only test the specified SDK as well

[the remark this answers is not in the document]
ajwootto: does that need to be a requirement? if so where do we enforce that?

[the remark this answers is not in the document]
ajwootto: that makes sense but we should figure out a way to make it "impossible" to trigger this with the GITHUB_SHA set without specifying the SDK to test as well

proxies/nodejs/install.sh
jsalaber: looks like it's using those packages from NPM instead: ```    "@devcycle/bucketing-assembly-script": "^1.1.20",
    "@devcycle/types": "1.0.52",```

ajwootto: you'll have to add each of those as well then

harness/features/variable.cloud.test.ts
```diff
@@ -19,7 +19,7 @@ const scope = getServerScope()
 describe('Variable Tests - Cloud', () => {
     forEachSDK((name) => {
         let url: string
-        const capabilities: string[] = SDKCapabilities[name]
+        const capabilities: string[] = SDKCapabilities[name] || []
```
ajwootto, line 22: this shouldnt need a default it would be an error for it to not find capabilities in here

ajwootto (approved):

PR #40, opened 2022-12-06
cleanup parse location

ajwootto:
- some cleanup to the parse location function to make it clearer how the URL is decomposed


suthar26 (approved):
looks good

PR #41, opened 2022-12-06
fail test if nock has pending interceptors, or a request was made that was not matched

ajwootto:
- add a standard "afterEach" block that checks that nock has no pending mocks
- also check that no unmatched requests were made (ie. nock provided a default response because no mock was defined)
- also clean up and reset nock between tests


suthar26 (approved):

harness/features/initialize.cloud.test.ts
```diff
@@ -1,8 +1,8 @@
 import { getConnectionStringForProxy, forEachSDK, describeIf, createClient, createUser, wait } from '../helpers'
 import { Capabilities, SDKCapabilities } from '../types'
 import { v4 as uuidv4 } from 'uuid'
-import { getServerScope } from '../mockServer'
 import nock from 'nock'
+import { getServerScope } from '../nock'
```
suthar26, line 5: linting will throw error (or alteast it does in ttroi) for multiple before singular import but thats meh

ajwootto: doesn't seem like the lint rules in here care about that (assuming linting is running)

suthar26: yeah we dont have them running here

jsalaber (approved):

PR #47, opened 2022-12-07
fix: wrong condition for creating .env file

jsalaber:


elliotCamblor (approved):

.github/workflows/config.yml
```diff
@@ -27,10 +27,10 @@ jobs:
           fetch-depth: 0
       - name: Write variables to .env file
         run: |
-          if [ -f ".env" ]
+          if [ ! -f ".env" ]
```
ajwootto, line 30: is this how you do negation?

ajwootto: yeah but i mean is this valid bash?

jsalaber: i believe so, i've tested the if statement locally

PR #56, opened 2022-12-08
Add cloud test client helper

ajwootto:
Separate the cloud and local clients into two separate test helpers

Both helpers pass the appropriate option to the proxy when calling `createClient` to make the right kind of client

The call methods (callVariable, callAllVariables etc.) correctly set the `isAsync` property to true or false depending on the client.

Also fixed some sources of errors happening on the proxy server:
- initialize tests now respond to CDN request with a valid config
- event API mocking now always returns 201 instead of 200

The "wait for initialize" method will now make sure the request was successful and that the SDK did not throw an error. The same will happen for track unless you specify that you're expecting it to fail.

### DIFF
--- a/harness/helpers.ts
+++ b/harness/helpers.ts
@@ -29,9 +29,6 @@ export const getConnectionStringForProxy = (proxy: string) => {
     return `http://${host}:${port}`
 }
 
-export const mockServerUrl = `http://${process.env.DOCKER_HOST_IP
-    ?? 'host.docker.internal'}:${global.__MOCK_SERVER_PORT__}`
-
 export const forEachSDK = (tests) => {
     // get the list of SDK's and their capabilities
     let SDKs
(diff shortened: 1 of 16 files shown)


harness/helpers.ts
```diff
@@ -277,3 +239,49 @@ export class TestClient {
...
+            this.clientId,
+            this.sdkKey,
+            {
+                baseURLOverride: `${getMockServerUrl()}/client/${this.clientId}`,
+                enableCloudBucketing: true,
+                ...options
+            }
+        )
+
+        this.clientLocation = response.headers.get('location')
```
nsalamad, line 259: do we still need to respond with the location header from the `/client` endpoint? The test harness is the one that chooses the clientId so it should just know the location i.e. `/client/${clientId}`

ajwootto: The test harness knows the clientId but it shouldn't assume it knows the location. Some of the other code here is assuming that the client path is `/client/${clientId}` when it should be respecting the location header from the proxy

nsalamad: That's a valid point. But also some of the tests kind of need to assume what the location is for the mocked endpoints - the config endpoints needs to be mocked before the client is created otherwise the sdk is gonna receive a 404 for its first fetch. It's a little weird but I agree that the harness shouldn't assume the location

ajwootto: the config endpoint is the concern of the harness and isn't really related to the "location" of the client on the proxy. That location is owned by the proxy

PR #57, opened 2022-12-08
override global fetch to log full exception

ajwootto:
- override the global fetch function to always log the full error object as well as throw it


elliotCamblor (approved):
oldFetch is where you throw a ball for an old dog and they dont chase the ball

ajwootto:
> oldFetch is where you throw a ball for an old dog and they dont chase the ball

okay

PR #60, opened 2022-12-12
docs: add OpenAPI spec

nsalamad:
# Changes
* Added spec.yaml - OpenAPI spec of  proxy endpoints
* Added proxies/README.md - Markdown version of our spec
* Added widdershins templates - Used to customize the look of proxies/README.md


proxies/README.md
```diff
@@ -0,0 +1,329 @@
+
+<h1 id="devcycle-test-harness-proxy">DevCycle Test Harness - Proxy v1.0.0</h1>
+
+# Overview
+The SDK proxy server is responsible for wrapping the SDK under test with an HTTP interface that can be used to call any of its public-facing methods with an arbitrary set of arguments,
+and returning the response to the caller (which will be the harness server). It will also allow configuration options to be passed which override the base API URL the SDK makes requests to.
+Setting that URL to the test harness server’s address allows the harness to mock the backend APIs to return data back (or to trigger errors).
+
+### About `entityType`
+To check that the SDK methods are returning a correct type of an instance, they will all contain an `entityType` in their response. `enityType` can be one of four values:
```
ajwootto, line 10: I think Parth also added "Void"

proxies/widdershins/options.json
```diff
@@ -0,0 +1,7 @@
+{
```
ajwootto, line 1: it might be better if all of this (including the spec etc.) went in a `docs/` folder at the root of the repo. It's a bit confusing to have a bunch of folders in here named after different SDKs then just have `widdershins` in there as well

proxies/README.md
```diff
@@ -0,0 +1,329 @@
...
+
+`POST /client`
+
+*Create a DVC client instance*
+
+> Body parameter
+
+```json
+{
+  "clientId": "string",
```
ajwootto, line 27: any way to add descriptions for the individual request body fields?

nsalamad: this is what I was referring to in the slack about it not being perfect. I spent some time this morning trying to get the descriptions for request bodies to show up in the markdown file but for some reason it doesn't. I'm thinking the spec might be wrong but when I use redocly to preview, everything looks good so I'm thinking it's something with widdershins. I'll dig into it a bit more but idk if it's worth spending more time on it

proxies/docs/README.md
```diff
@@ -0,0 +1,349 @@
+
+<h1 id="devcycle-test-harness-proxy">DevCycle Test Harness - Proxy v1.0.0</h1>
```
ajwootto, line 2: i was actually thinking this could be a top-level docs folder rather than being inside "proxies"

docs/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Response Headers
+
+|Status|Header|Type|Format|Description|
+|---|---|---|---|---|
+|201|Location|string||Location of the Client instance on the proxy server|
+
+<a id="opIdcommand"></a>
+
+`POST /client/{clientId}/command`
```
ajwootto, line 96: Is this URL correct? Pretty sure it currently does not have the "command" at the end

docs/README.md
```diff
@@ -0,0 +1,349 @@
...
+```
+
+<h3 id="command-parameters">Parameters</h3>
+
+|Name|In|Type|Required|Description|
+|---|---|---|---|---|
+|`command`|body|string|false|The name of the SDK method to invoke|
+|`isAsync`|body|boolean|false|Tells the proxy weather to wait on the SDK method or not|
+|`params`|body|[anyOf]|false|Arguments to the  SDK method|
+|`» *anonymous*`|body|[LocationParam](#schemalocationparam)|false|An argument to an SDK method|
```
ajwootto, line 127: whats with this *anonymous* stuff

docs/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+> Created
+
+```json
+{
+  "message": "Success"
+}
+```
+
+```json
```
ajwootto, line 53: in the error case (where the SDK threw an error) is the status code of the response the same and its just the message that tells us about the error? Should this be "exception" instead?

docs/README.md
```diff
@@ -0,0 +1,349 @@
...
+    "_id": "638681f059f1b81cc9e6c7fa",
+    "key": "bool-var",
+    "type": "Boolean",
+    "value": true
+  },
+  "logs": []
+}
+```
+
+```json
```
ajwootto, line 150: this should document the "errors from SDK" responses, like `exception` and `asyncError`

docs/README.md
nsalamad: I thought we wanted to add "command" at the end. You brought it up on slack so I assumed we'd make the change. I'll remove it for now if that's no longer the plan

docs/README.md
nsalamad: yeah the status code should be 200 but the body should contain `exception` instead. I'll make the change

docs/README.md
nsalamad: It's how widdershins handles listing the anyOf items. I agree, It's kind of janky but the other option is to use the --shallowSchemas option which just removes LocationParam and ValueParam from the table. I could also dig through the widdershins code and see if we can get more desired outcome

docs/README.md
nsalamad: ok i'll make a ticket to change the endpoint

docs/README.md
```diff
@@ -0,0 +1,368 @@
...
+|400|[Bad Request](https://tools.ietf.org/html/rfc7231#section-6.5.1)|Bad Request|
+
+<h3 id="createclient-responseschema">Response Schema</h3>
+
+Status Code **201**
+
+|Name|Type|Required|Description|
+|---|---|---|---|
+|`» message`|string|false|A message indicating the result of initializing an SDK client|
+|`» exception`|string|false|An error from the SDK indicating an exception was thrown during initialization|
```
ajwootto, line 81: the exception status code is 200 as far as I know

docs/README.md
```diff
@@ -0,0 +1,368 @@
...
+and returning the response to the caller (which will be the harness server). It will also allow configuration options to be passed which override the base API URL the SDK makes requests to.
+Setting that URL to the test harness server’s address allows the harness to mock the backend APIs to return data back (or to trigger errors).
+
+### About `entityType`
+To check that the SDK methods are returning a correct type of an instance, they will all contain an `entityType` in their response. `enityType` can be one of four values:
+  `Variable`, `User`, `Feature`, `Object`, `Void`
+
+<h1 id="devcycle-test-harness-proxy-client">Client</h1>
+
+SDK Client
```
ajwootto, line 15: would be good to add a short description to each endpoint describing what its for

ajwootto: I guess there is one actually I just missed it. Maybe what's missing is a general sense of how these endpoints are used in conjunction?

docs/README.md
```diff
@@ -0,0 +1,368 @@
...
+
+### Response Headers
+
+|Status|Header|Type|Format|Description|
+|---|---|---|---|---|
+|201|Location|string||Location of the Client instance on the proxy server|
+
+<a id="opIdcommand"></a>
+
+`POST /client/{clientId}/command`
```
ajwootto, line 97: this endpoint appears to be in the client section, but technically its a general endpoint for calling commands on anything based on its location header (that way you can also call methods on variables etc.)

The actual path for this would be something like 
`/:location/command`

nsalamad: yeah, the only reason I decided not to do that is cause I thought that'd be too generalized. I thought it would make more sense to have individual 'command' sections for other instances that you can also call methods on. Personally I think it's better to be more specific when writing docs but I can generalize it and make it `/:location/command`, it would save us time later down the line since we wouldn't need to update the spec

docs/README.md
nsalamad: I've added a "How to use" section to the overview. lmk if it looks good

docs/README.md
```diff
@@ -0,0 +1,388 @@
...
+{
+  "asyncError": "Must have a user_id set on the user"
+}
+```
+
+<h3 id="command-responses">Responses</h3>
+
+|Status|Meaning|Description|
+|---|---|---|
+|200|[OK](https://tools.ietf.org/html/rfc7231#section-6.3.1)|OK|
```
ajwootto, line 289: I think we changed the response code here to be 201 when the command was successful (since it creates a "command result")

It's still 200 for the cases where there's an exception or asyncError

docs/README.md
```diff
@@ -0,0 +1,388 @@
...
+
+|Name|Type|Required|Description|
+|---|---|---|---|
+|`» message`|string|false|A message indicating the result of initializing an SDK client|
+
+Status Code **400**
+
+|Name|Type|Required|Description|
+|---|---|---|---|
+|`» error`|string|true|An error message indicating what went wrong with the request|
```
ajwootto, line 103: this field is called "message" still

nsalamad: looking at the implementation, when we return a 400, it has `error` instead of `message`

nsalamad: oh nvm I was looking at the wrong thing. Will fix

docs/README.md
ajwootto: the reason it's important to say its for any "location" is that the design is specifically made so that anything which has been identified with a location header can have commands called on it. The spec doesn't actually define what the "location" format is supposed to be since its the concern of the proxy server how to describe where something is basically, and the harness is supposed to just use whatever the proxy says the location is.

nsalamad: yup, makes sense

docs/spec.yaml
```diff
@@ -0,0 +1,357 @@
...
+                    description: 'Type of entity returned. One of `Variable`, `User`, `Feature`, `Object` or `Void`'
+                  data:
+                    type: object
+                    description: 'Value returned by the method'
+                  logs:
+                    type: array
+                    description: 'Logs generated during the method invocation'
+                    items:
+                      type: string
+                  exception:
```
ajwootto, line 135: this should be under a "200" response. It's only 201 if the command succeeded

nsalamad: yup, made the change

ajwootto (approved):

PR #63, opened 2022-12-12
feat: user endpoint dotnet

jsalaber:
# Summary

Implement `user` endpoint for the dotnet proxy server.

### DIFF
--- a/proxies/dotnet/Controllers/User.cs
+++ b/proxies/dotnet/Controllers/User.cs
@@ -1,0 +1,56 @@
+using System.Net;
+using Microsoft.AspNetCore.Mvc;
+using DevCycle.SDK.Server.Common.Model;
+
+// using Newtonsoft.Json;
+using System.Text.Json;
+
+namespace dotnet.Controllers;
+
+[ApiController]
+[Route("[controller]")]
+public class UserController : ControllerBase
+{
+
+    private readonly ILogger<UserController> _logger;
+
+    public UserController(ILogger<UserController> logger)
+    {
+        _logger = logger;
+    }
+
+    [HttpPost]
+    public object CreateUser(ClientRequestUser user)
+    {
+        try {
+            var sdkUser = new User(
+                user.UserId, 
+                user.Email,
+                user.Name,
+                user.Language,
+                user.Country,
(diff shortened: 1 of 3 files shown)


proxies/dotnet/Controllers/User.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    public UserController(ILogger<UserController> logger)
+    {
+        _logger = logger;
+    }
+
+    [HttpPost]
+    public async Task<IActionResult> CreateUser()
+    {
+        var body = await Request.GetRawBodyAsync().ConfigureAwait(false);
```
ajwootto, line 29: couldn't you just use the User class from the SDK here as an argument to `CreateUser` rather than manually reading the body like this?

jsalaber: I wasn't able to use the `DVCPopulatedUser` class unfortunately so I had to make its own model that could be used with model binding

proxies/dotnet/Controllers/User.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        var userId = users.Count;
+        users[userId.ToString()] = user;
+
+        HttpContext.Response.Headers.Add("Location", "user/" + userId);
+
+        var res = new ContentResult();
+        res.ContentType = "application/json";
+        res.StatusCode = (int)HttpStatusCode.Created;
+        res.Content = JsonSerializer.Serialize(result);
```
ajwootto, line 37: you should be able to just return result here rather than these four lines

jsalaber: part of the ticket is to have `201` returned since it's creating a new resource

ajwootto: does it not do 201 by default on POST?

even if not, you only then need the line that sets the status code. The content type header and JSON body are set automatically when you return a serializable object

jsalaber: it does `200` by default, but I changed it to just change the response status code like so: 
```
Response.StatusCode = (int)HttpStatusCode.Created;
```

ajwootto (approved):

PR #64, opened 2022-12-12
DVC-5431 client endpoint dotnet

ajwootto:
- update SDK versions to resolve incompatible versions
- write /client endpoint to init cloud vs. local bucketing SDK
- still getting issues with wasmtime working locally and on the docker image...

### DIFF
--- a/proxies/dotnet/Controllers/Client.cs
+++ b/proxies/dotnet/Controllers/Client.cs
@@ -1,14 +1,21 @@
 using Microsoft.AspNetCore.Mvc;
 using DevCycle.SDK.Server.Cloud.Api;
 using DevCycle.SDK.Server.Local.Api;
-using DevCycle.SDK.Server.Common.Model;
-using DevCycle.SDK.Server.Common;
+using DevCycle.SDK.Server.Common.Model.Local;
+using DevCycle.SDK.Server.Common.Model.Cloud;
+using System.Text.Json.Serialization;
 
 namespace dotnet.Controllers;
 
+public class AllOptions: DVCLocalOptions {
+    [JsonPropertyName("enableEdgeDB")]
+    public bool EnableEdgeDB { get; set; }
+}
+
 public class ClientRequestBody {
     public string clientId { get; set;}
-    public Dictionary<string, object> options { get; set;}
+    public AllOptions options { get; set;}
+
     public string? sdkKey { get; set;}
 
     public bool? cloudBucketing { get; set; }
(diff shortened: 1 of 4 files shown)


proxies/dotnet/Controllers/Client.cs
```diff
@@ -0,0 +1,90 @@
...
+
+                if (clientBody.options != null) {
+                    cloudOptions = new DVCCloudOptions(enableEdgeDB: clientBody.options.EnableEdgeDB ?? false);
+                }
+
+                DataStore.CloudClients[clientBody.clientId] = new DVCCloudClientBuilder()
+                    .SetEnvironmentKey(clientBody.sdkKey)
+                    .SetOptions(cloudOptions)
+                    .SetLogger(new LoggerFactory())
+                    .Build();
```
emmawillis, line 67: will this set the BaseURLOverride on the cloud client? i thought we'd need to set restClientOptions with the baseUrl here

proxies/dotnet/Controllers/Client.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        [JsonPropertyName("baseURLOverride")]
+        public string? BaseURLOverride { get; set; }
+
+        [JsonPropertyName("enableEdgeDB")]
+        public bool? EnableEdgeDB { get; set; }
+    }
+
+public class ClientRequestBody {
+    public string? clientId { get; set;}
```
jsalaber, line 26: Pascal case for properties?

proxies/dotnet/Controllers/Client.cs
```diff
@@ -0,0 +1,97 @@
...
+        new public int EventFlushIntervalMs { get; set; }
+
+        [JsonPropertyName("baseURLOverride")]
+        public string? BaseURLOverride { get; set; }
+
+        [JsonPropertyName("enableEdgeDB")]
+        public bool? EnableEdgeDB { get; set; }
+    }
+
+public class ClientRequestBody {
```
jsalaber, line 25: should we put these in a models folder

jsalaber (changes requested):

proxies/dotnet/Controllers/Client.cs
ajwootto: I could but then I'd have to specify the real name with the JSON property override and this just seemed easier 😬

proxies/dotnet/Controllers/Client.cs
ajwootto: personally I prefer it to be colocated with the place its actually going to be used since its not really a "model" thats used anywhere else, it's specifically the body of this one request

jsalaber: it might be convenient for this file since it's a small request body but for e.g. my pr where the user request can be a big request body, it made more sense for me to separate it to have the file more readable, so would I then put it in the file even if it makes it larger / harder to read?

ajwootto: I guess in that case its fine to have it separate, like our DTOs in Nest.

However in Nest we still locate the files defining those DTOs within the same "module" so that theyre easy to find relative to the other files that define that set of endpoints.

proxies/dotnet/Controllers/Client.cs
jsalaber: it's a little odd that you've defined the ClientOptions using `JsonPropertyName` but not the top level RequestBody

ajwootto: thats partly cuz the ClientOptions need to match the interface of the SDK options object

ajwootto: I can change these for consistency

jsalaber (approved):

PR #66, opened 2022-12-13
Set up java scaffolding

elliotCamblor:


docker-compose.yml
```diff
@@ -28,3 +28,11 @@ services:
...
     ports:
       - "3000"
+  java:
+    build:
+      context: ./proxies/java
+      args:
+        - JAVA_SDK_VERSION=${JAVA_SDK_VERSION}
+    container_name: java
+    ports:
+      - "8080"
```
jsalaber, line 38: why `8080`?

jsalaber (changes requested):

proxies/java/src/main/java/com/dvc_harness/proxy/data/DataStore.java
```diff
@@ -0,0 +1,23 @@
...
+import com.devcycle.sdk.server.common.model.*;
+
+class DataStore {
+    public static Hashtable<String, DVCLocalClient> LocalClients;
+    public static Hashtable<String, DVCCloudClient> CloudClients;
+
+    public static Hashtable<String, User> Users;
+    public static Hashtable<String, String> Commands;
+
+    public DataStore()
```
ajwootto, line 16: is this a constructor for an instance of this? don't you want to use the static properties instead? This constructor would then need to be static

docker-compose.yml
elliotCamblor: couldnt figure out how to get 3000 to work

proxies/java/src/main/java/com/dvc_harness/proxy/data/DataStore.java
elliotCamblor: this was a complete afterthought on the rest of the PR

elliotCamblor: just threw it together because I saw it in the c# scaffolding, will need to figure out how this should actually look

ajwootto: might be better to just remove it then if this isn't intended to be used as-is, lest somebody try and use it

elliotCamblor: yea I think just making it static should do what we want

docker-compose.yml
elliotCamblor: its working on 3000 now

proxies/java/src/main/java/com/dvc_harness/proxy/data/DataStore.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public static Hashtable<String, User> Users;
+    public static Hashtable<String, String> Commands;
+
+    public static void DataStore()
+    {
+        Hashtable CloudClients = new Hashtable<String, DVCCloudClient>();
+        Hashtable LocalClients = new Hashtable<String, DVCLocalClient>();
+        Hashtable Users = new Hashtable<String, User>();
+        Hashtable Commands = new Hashtable<String, Object>();
```
ajwootto, line 21: this disagrees with the type definition of the `Commands` declaration above?

ajwootto: also are you sure these are talking about the same property? You shouldn't need to re-declare the type being `Hashtable` for these assignments

elliotCamblor: hmmm I really need to enable autosave in vscode so I stop committing incomplete changes

elliotCamblor: ah I see what you mean

jsalaber (approved):

PR #70, opened 2022-12-15
add client endpoint for java proxy

ajwootto:
- add `/client` endpoint for creating new Java clients

### DIFF
--- a/proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
+++ b/proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
@@ -1,11 +1,19 @@
 package com.dvc_harness.proxy.controller;
 
-import org.springframework.beans.factory.annotation.Qualifier;
-import org.springframework.beans.factory.annotation.Value;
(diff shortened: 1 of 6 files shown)


proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
```diff
@@ -21,4 +29,41 @@ public Spec spec() {
         return new Spec();
     }
 
+    @PostMapping("/client")
+    public MessageResponse client(@RequestBody ClientRequestBody body, HttpServletResponse response) {
+        if (body.enableCloudBucketing != null && body.enableCloudBucketing) {
```
elliotCamblor, line 34: whats up with this check? Why check if its not null and then check if its truthy?

elliotCamblor (approved):

proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
elliotCamblor: lovely

PR #74, opened 2022-12-16
java client error handling

ajwootto:


proxies/java/src/main/java/com/dvc_harness/proxy/models/AsyncErrorResponse.java
```diff
@@ -0,0 +1,9 @@
+package com.dvc_harness.proxy.models;
+
+public class AsyncErrorResponse extends BaseResponse {
+    public final String asyncError;
+
+    public AsyncErrorResponse(String message) {
```
elliotCamblor, line 6: do you know if these classes need getters? I know when I was setting up the spec endpoint spring was unhappy when I tried to return class instances that didnt have getters

ajwootto: it doesnt seem to need one (for the messageresponse anyways). I get the serialized response back just fine

elliotCamblor: okay, cool

elliotCamblor (approved):

PR #75, opened 2022-12-16
Dvc 5462 user endpoint

elliotCamblor:
I only added the fields that were actively using to the builder in the endpoint. I can add the rest if we want, it just gets unruly

### DIFF
--- a/proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
+++ b/proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
@@ -11,8 +11,14 @@ import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RestController;
 
+
+import com.dvc_harness.proxy.models.ClientRequestBody;
+import com.dvc_harness.proxy.models.UserRequestBody;
+import com.dvc_harness.proxy.models.MessageResponse;
+
 import com.devcycle.sdk.server.local.api.DVCLocalClient;
 import com.devcycle.sdk.server.cloud.api.DVCCloudClient;
+import com.devcycle.sdk.server.common.model.User;
 
 @Controller
 @RestController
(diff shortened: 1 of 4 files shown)


proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
```diff
@@ -28,6 +30,29 @@ public Spec spec() {
         return new Spec();
     }
 
+    @PostMapping("/user")
+    public BaseResponse user(@RequestBody UserRequestBody body, HttpServletResponse response) {
+        try {
+            User sdkUser = User.builder()
+                .userId(body.user_id)
+                .name(body.name)
```
ajwootto, line 38: does this work when these fields are null / aren't set?

proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
```diff
@@ -28,6 +30,29 @@ public Spec spec() {
...
+                .customData(body.customData)
```
ajwootto, line 39: i think you might as well just list every property here

proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
elliotCamblor: yep

proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
elliotCamblor: but its so ugly 😭 okay tho

ajwootto: thats Java babyyyy

ajwootto: you could probably write some fancy reflection thing to pull the correct method name based on the name of each property in the request

elliotCamblor: id rather just get it done lol

proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
```diff
@@ -28,6 +30,36 @@ public Spec spec() {
...
+                .appVersion(body.appVersion)
+                .appBuild(body.appBuild)
+                .customData(body.customData)
+                .privateCustomData(body.privateCustomData)
+                .createdDate(body.createdDate)
+                .lastSeenDate(body.lastSeenDate)
+                .build();
+
+            var userId = DataStore.Users.size();
+            DataStore.Users.put(userId + "", sdkUser);
```
nsalamad, line 50: Might be more clear to do `var userId = Interger.toString(DataStore.Users.size());`

elliotCamblor: 👍

nsalamad (approved):

proxies/java/src/main/java/com/dvc_harness/proxy/controller/ProxyController.java
```diff
@@ -28,6 +30,36 @@ public Spec spec() {
...
+                .lastSeenDate(body.lastSeenDate)
+                .build();
+
+            var userId = Integer.toString(DataStore.Users.size());
+            DataStore.Users.put(userId, sdkUser);
+
+            var result = new UserData(sdkUser);
+
+            response.addHeader("Location", "user/" + userId);
+            return result;
```
ajwootto, line 55: is the status here 201?

ajwootto (approved):

PR #76, opened 2022-12-16
Add location to dotnet proxy

emmawillis:


proxies/dotnet/Controllers/Location.cs
```diff
@@ -0,0 +1,189 @@
...
+        } catch (Exception e) {
+            Response.StatusCode = 404;
+            return new { message = e.Message };
+        }
+
+        try {
+            var result = await InvokeCommand(entity, body.Command, body.IsAsync, parsedParams);
+
+            Response.StatusCode = 201;
+            if (result.CommandId != null) Response.Headers.Add("Location", $"command/{body.Command}/{result.CommandId}");
```
ajwootto, line 78: when would this commandID be null?

proxies/dotnet/Controllers/Location.cs
```diff
@@ -0,0 +1,189 @@
...
+
+    private string GetEntityType(object result) {
+        var type = TypeDescriptor.GetClassName(result) ?? "Object";
+        if (type.Contains("Dictionary")) {
+            type = "Object";
+        } else {
+            type = type.Split(".").Last();
+        }
+
+        return EntityTypes.Contains(type) ? type : "Void";
```
ajwootto, line 187: I think "Void" is supposed to represent "nothing was returned" not "we dont know what type this corresponds to". That case should be "Object"

proxies/dotnet/Controllers/Location.cs
```diff
@@ -0,0 +1,189 @@
...
+
+    private List<object> ParseParams(List<JToken> bodyParams) {
+        var result = new List<object>{};
+
+        foreach (var param in bodyParams) {
+            if (param["location"] != null) {
+                result.Add(GetEntity(param["location"].ToString()));
+                Console.WriteLine(Newtonsoft.Json.JsonConvert.SerializeObject(GetEntity(param["location"].ToString())));
+            } else if (param["value"] != null) {
+                if (param["value"].Type == JTokenType.Float) {
```
ajwootto, line 111: is this kinda like a generic JSON cast of the data to a native type?

proxies/dotnet/Controllers/Location.cs
elliotCamblor: yeah, really not sure why this is being done. I think it shouldnt be object either tbh, object should already be covered. This method shoulld probably just return type

proxies/dotnet/Controllers/Location.cs
elliotCamblor: yeah, this is what we spent all afternoon trying to get working yesterday

ajwootto: nice

proxies/dotnet/Controllers/Location.cs
elliotCamblor: I dont think it would be

proxies/dotnet/Controllers/Location.cs
elliotCamblor: im just gonna make it object for now to appease the enum gods

PR #78, opened 2022-12-19
feat: install dotnet local package if github sha is present

jsalaber:
# Summary

Added `SDK_GITHUB_SHA` to docker image, install local package when it's present.


proxies/dotnet/install.sh
```diff
@@ -1,13 +1,36 @@
+if [ -n "$SDK_GITHUB_SHA" ]; then
```
ajwootto, line 1: werent you also checking the SDKS_TO_TEST variable on the other platforms?

proxies/dotnet/install.sh
```diff
@@ -1,13 +1,36 @@
...
+    # changing directory so dotnet doesn't run again when starting up the server
+    cd ..
+    git clone https://github.com/DevCycleHQ/dotnet-server-sdk.git
```
ajwootto, line 4: shouldnt this be using the github sha?

jsalaber: oh yes my mistake

ajwootto: bump

ajwootto: oh nvm

ajwootto (approved):

PR #83, opened 2022-12-21
Split base URL override to match new changes

JamieSinn:


harness/helpers.ts
```diff
@@ -291,7 +291,12 @@ export class LocalTestClient extends BaseTestClient {
             false,
             this.clientId,
             this.sdkKey,
-            { baseURLOverride: `${getMockServerUrl()}/client/${this.clientId}`, ...options }
+            {
+                eventsAPIBaseURL: `${getMockServerUrl()}/client/${this.clientId}`,
```
ajwootto, line 295: are you changing the names of these properties on C# to match?

JamieSinn: If these are the ones we go with I will

ajwootto (approved):

PR #89, opened 2022-12-22
Update location endpoint to new format c#

elliotCamblor:


proxies/dotnet/Controllers/Location.cs
```diff
@@ -100,13 +104,39 @@ public async Task<object> Post([FromBody] PostBody body, string location)
...
 
-    private List<object> ParseParams(List<JToken> bodyParams) {
+    private List<object> ParseParams(List<JToken> bodyParams, ClientRequestUser user, ClientRequestEvent reqEvent) {
         var result = new List<object>{};
 
         foreach (var param in bodyParams) {
-            if (param["location"] != null) {
-                result.Add(GetEntity(param["location"].ToString()));
-                Console.WriteLine(Newtonsoft.Json.JsonConvert.SerializeObject(GetEntity(param["location"].ToString())));
+            if (param["type"] != null && param["type"].Value<string>() == "user") {
```
jsalaber, line 111: I think you can do `param["type"]?.Value` to avoid the extra null check

jsalaber (changes requested):

jsalaber (approved):

PR #91, opened 2022-12-22
chore: added helper for comparing exception messages

jsalaber:
# Summary

Added exception message helper method to check if exception message exists.

### DIFF
--- a/harness/features/initialize.cloud.test.ts
+++ b/harness/features/initialize.cloud.test.ts
@@ -2,7 +2,7 @@ import {
     getConnectionStringForProxy,
     forEachSDK,
     describeIf,
-    CloudTestClient, describeCapability
+    CloudTestClient, describeCapability, expectErrorMessageToBe
 } from '../helpers'
 import { Capabilities, SDKCapabilities } from '../types'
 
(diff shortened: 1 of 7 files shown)


harness/features/initialize.cloud.test.ts
```diff
@@ -34,7 +37,10 @@ describe('Client Initialize Tests - Cloud', () => {
                 const client = new CloudTestClient(name)
                 const response = await client.createClient({}, null, true)
                 const body = await response.json()
-                expect(body.exception).toBe('Missing environment key! Call initialize with a valid environment key')
+                expectExceptionMessageToBe(
+                    'Missing environment key! Call initialize with a valid environment key',
```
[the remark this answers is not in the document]
jsalaber: I thought it would make more sense in the context of how I named the function, but I can change the name of the function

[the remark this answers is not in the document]
ajwootto: i think even with this name it makes more sense the other way, since you're saying `expect <exception message> to be <value>`
not `expect <value> to be <exception message>`

harness/features/initialize.cloud.test.ts
```diff
@@ -25,7 +25,10 @@ describe('Client Initialize Tests - Cloud', () => {
                 const client = new CloudTestClient(name)
                 const response = await client.createClient({}, 'invalidKey', true)
                 const body = await response.json()
-                expect(body.exception).toBe('Invalid environment key provided. Please call initialize with a valid server environment key')
+                expectErrorMessageToBeTheSame(
```
ajwootto, line 28: maybe just `expectErrorMessageToBe`?

jsalaber: with that wording it makes me think there's only going to be one param tho

ajwootto: I dont read it that way, to me its just the same as when you write out the `expect` statement

ajwootto (approved):

PR #102, opened 2023-01-03
Test user platform on event tests

emmawillis:


harness/helpers.ts
```diff
@@ -430,3 +430,7 @@ export const expectErrorMessageToBe = (message: string, expected: string) => {
     // the exception and the expected exception message
     expect(message).toBeTruthy()
 }
+
+export const getPlatformBySdkName = (name: string) => {
+    return name === 'DotNet' ? 'C#' : name // TODO use Sdks.dotnet instead of 'DotNet' when enable dotnet
```
ajwootto, line 435: I think DotNet has a distinction between Cloud and Local for the platform..

emmawillis: oh wait you're right, it shows up as just C# in snowflake but the actual event has cloud/local as well

ajwootto: hmm how come it shows up that way in Snowflake then?

emmawillis: not sure let me check again

emmawillis: ah never mind the events i was looking at were from May, i'm assuming before we added local as an option - the recent ones have the right format

ajwootto (approved):

PR #108, opened 2023-01-04
test config fetching with etag and failure behaviour

ajwootto:
- test that the cached config is used when the reply is 304 not modified
- test that the cached config is used when the reply is a failure, but that it retries that request once
- test that the cached config used when the reply is not valid JSON

### DIFF
--- a/harness/features/initialize.local.test.ts
+++ b/harness/features/initialize.local.test.ts
@@ -110,7 +110,7 @@ describe('Initialize Tests - Local', () => {
                 await testClient.close()
             })
 
-            it('uses the same config if the refetch fails', async () => {
+            it('uses the same config if the refetch fails, after retrying once', async () => {
                 const testClient = new LocalTestClient(name)
                 scope
                     .get(`/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`)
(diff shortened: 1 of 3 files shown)


harness/features/initialize.local.test.ts
```diff
@@ -84,6 +84,76 @@ describe('Initialize Tests - Local', () => {
...
+                    .reply(200, config)
+
+                scope
+                    .get(`/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`)
+                    .times(2)
+                    .reply(503, {})
+
+                await testClient.createClient(true, { configPollingIntervalMS: 1000 })
+
+                expect(scope.pendingMocks().length).toEqual(1)
```
elliotCamblor, line 126: does the `.times(2)` still just count as one pending mock?

ajwootto: yeah

nsalamad (approved):

harness/features/initialize.local.test.ts
```diff
@@ -84,6 +84,76 @@ describe('Initialize Tests - Local', () => {
...
+                scope
+                    .get(`/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`)
+                    .reply(200, config)
+
+                scope
+                    .get(`/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`)
+                    .times(2)
+                    .reply(503, {})
+
+                await testClient.createClient(true, { configPollingIntervalMS: 1000 })
```
nsalamad, line 124: maybe increase the polling interval so that we're sure the second time the mock endpoint is hit it's actually from the retry and not polling.

PR #111, opened 2023-01-04
enable the dotnet tests and make the test run not required

elliotCamblor:


.github/workflows/config.yml
```diff
@@ -16,6 +16,7 @@ jobs:
   run-tests:
     runs-on: ubuntu-latest
     name: Run Tests
+    continue-on-error: true
```
ajwootto, line 19: whats this for

elliotCamblor: thats supposed to make it optional I thought

elliotCamblor: nvm I misunderstood the docs

ajwootto: making it optional is just a repository setting

elliotCamblor: ah, I dont have permission which is why I cant see the setting

ajwootto (approved):

PR #117, opened 2023-01-05
use jest projects to run sdk tests in parallel

ajwootto:
Switch from running tests for each SDK using `describe.each` to using Jest "projects" which each define the SDK that should be tested.

TODO: figure out how to compartmentalize the docker compose files for each project to only run the appropriate proxy container.

### DIFF
--- a/harness/helpers.ts
+++ b/harness/helpers.ts
@@ -22,7 +22,6 @@ export const getMockServerUrl = () => {
     return `http://${process.env.DOCKER_HOST_IP ?? 'host.docker.internal'}:${global.__MOCK_SERVER_PORT__}`
 }
 
-
 export const getConnectionStringForProxy = (proxy: string) => {
     if (process.env.LOCAL_MODE === "1") {
         return `http://localhost:3000`
@@ -45,7 +44,7 @@ const clientTestNameMap: Record<string, string> = {}
 // It can be formatted as a JSON array or a comma-separated list.
 // The SDKs are returned as their human-readable names from the Sdks enum, not the enum keys.
 export const getSDKs = () => {
-    const SDKS_TO_TEST = process.env.SDKS_TO_TEST || global.JEST_PROJECT_SDK_TO_TEST
+    const SDKS_TO_TEST = process.env.SDKS_TO_TEST
 
     try {
         return JSON.parse(SDKS_TO_TEST ?? '').map((sdk) => Sdks[sdk])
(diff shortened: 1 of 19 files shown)


harness/helpers.ts
```diff
@@ -31,19 +31,25 @@ export const getConnectionStringForProxy = (proxy: string) => {
...
+        if (SDKS_TO_TEST && Sdks[SDKS_TO_TEST]) {
+            SDKs = [Sdks[SDKS_TO_TEST]]
         } else {
             console.log('No specified SDKs to test, running all tests')
             SDKs = Object.values(Sdks)
         }
     }
     const scope = getServerScope()
 
+    if (process.env.SDKS_TO_TEST && global.JEST_PROJECT_SDK_TO_TEST && process.env.SDKS_TO_TEST !== global.JEST_PROJECT_SDK_TO_TEST) {
```
elliotCamblor, line 48: isn't SDKS_TO_TEST an array? so wont this case always get hit

ajwootto: the environment variable is a string

ajwootto: oh but yes it would be unequal if its a stringified array

elliotCamblor: Oh it can be an array or a string, I forgot

elliotCamblor: as in, a single SDK not in a stringified array

ajwootto: yep

harness/helpers.ts
```diff
@@ -59,26 +62,31 @@ export const getSDKs = () => {
     }
 }
 
-export const forEachSDK = (tests) => {
+/**
+ * This is used within the jest tests to fetch the running SDK environment for the current jest project.
+ * See `jest.config.ts` for more details on all projects. We also use this function to get the current nock server scope
+ * and add an `afterEach()` method to cleanup the client.
+ */
+export const getSDKScope = (): { sdkName: string, scope: Scope } => {
```
jonathannorris, line 70: This is the main change here that is replacing `forEachSDK()`

jest.config.ts
```diff
@@ -1,11 +1,58 @@
...
     preset: '@trendyol/jest-testcontainers',
     transform: {
         '^.+\\.tsx?$': ['ts-jest', {}]
     },
     globalSetup: './jest-global.ts',
     testEnvironment: './jest-environment.ts',
     setupFilesAfterEnv: ['./jest-setup.js'],
+}
+
+const projects = [
```
jonathannorris, line 13: new Jest projects are defined here, that let us change the jest config on a per-project basis.

jonathannorris (approved):

PR #123, opened 2023-01-06
update the spec to match the new proxy endpoint behaviour

elliotCamblor:


docs/spec.yaml
```diff
@@ -282,6 +233,10 @@ components:
               command:
                 type: string
                 description: 'The name of the SDK method to invoke'
+              user:
+                $ref: '#/components/schemas/User'
+              Event:
```
ajwootto, line 238: should this be lowercase

proxies/nodejs/handlers/user.ts
```diff
@@ -1,26 +0,0 @@
-import Koa from 'koa'
```
ajwootto, line 1: if you're deleting this theres also a c# equivalent still isn't there?

elliotCamblor: theres not

elliotCamblor: but there is a HelloWorld controller that I deleted

ajwootto (approved):

PR #127, opened 2023-01-09
Fix allVariables failure case test

emmawillis:
this fixes the failing tests on https://github.com/DevCycleHQ/dotnet-server-sdk/pull/63

### DIFF
--- a/harness/features/allVariables.cloud.test.ts
+++ b/harness/features/allVariables.cloud.test.ts
@@ -2,6 +2,7 @@ import {
     getConnectionStringForProxy,
     forEachSDK,
     CloudTestClient, describeCapability,
+    expectErrorMessageToBe
 } from '../helpers'
 import { getServerScope } from '../nock'
 import { Capabilities } from '../types'
@@ -23,8 +24,9 @@ describe('allVariables Tests - Cloud', () => {
         describeCapability(name, Capabilities.cloud)(name, () => {
             it('should return an empty object if variables request fails', async () => {
                 scope
+                    .persist() // need to persist because the client will retry on 500
                     .post(`/client/${client.clientId}/v1/variables`)
-                    .reply(404)
+                    .reply(500)
 
                 const response = await client.callAllVariables({
                     user_id: 'test_user',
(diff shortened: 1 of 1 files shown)


harness/features/allVariables.cloud.test.ts
```diff
@@ -37,6 +38,20 @@ describe('allVariables Tests - Cloud', () => {
                 expect(variablesMap).toMatchObject({})
             })
 
+            it.failing('should throw if variables request fails on user error', async () => {
+                // this will fail until dotnet and node are released
```
jsalaber, line 42: will this test also pass if this ends up passing once the changes are released?

emmawillis: oh right yeah just checked and it fails if it passes - i guess i'll just mark it with skip for now

ajwootto: it's better to just merge the failing test in this repo I think. We still haven't figured out the full process of how to write tests over here and then fix them over in the SDK repos, but I think it's better to have failing tests in here that then have to pass on the SDK PR

emmawillis: ah true that makes sense

harness/features/allVariables.cloud.test.ts
```diff
@@ -37,6 +38,19 @@ describe('allVariables Tests - Cloud', () => {
...
+                scope
+                    .post(`/client/${client.clientId}/v1/variables`)
+                    .reply(401, { message: 'Invalid sdk token' })
+
+                const response = await client.callAllVariables({
+                    user_id: 'test_user',
+                    email: 'user@gmail.com'
+                }, true)
+                const res = await response.json()
+                expect(res.asyncError).toContain('Invalid sdk token')
```
elliotCamblor, line 51: you may want to use the `expectErrorMessageToBeTruthy` helper function in case any SDK doesnt have the exact same error message

emmawillis: ah true, i was trying to get around that by using toContain since in theory they should always include the message body somewhere in the exception (at least dotnet and node both do) - do we want to both asserting on that, or just that it throws at all?

elliotCamblor: I think for consistency its fine to assert that it throws at all. If we ever undertake an initiative to ensure that all error messages are the same then we can just update the `expectErrorMessageToBeTruthy` function in one place and be done with it

harness/features/allVariables.cloud.test.ts
```diff
@@ -37,6 +38,19 @@ describe('allVariables Tests - Cloud', () => {
...
+                scope
+                    .post(`/client/${client.clientId}/v1/variables`)
+                    .reply(401, { message: 'Invalid sdk token' })
+
+                const response = await client.callAllVariables({
+                    user_id: 'test_user',
+                    email: 'user@gmail.com'
+                }, true)
+                const res = await response.json()
+                expectErrorMessageToBe(res, 'Invalid sdk token')
```
elliotCamblor, line 51: I think youll need to more specifically pass in `res.asyncError`

emmawillis: good catch, fixed!

harness/features/allVariables.cloud.test.ts
emmawillis: makes sense, fixed!

elliotCamblor (approved):

PR #139, opened 2023-01-12
client endpoint

elliotCamblor:


proxies/go/handler_client.go
```diff
@@ -0,0 +1,46 @@
...
+    WaitForInitialization bool `json:"waitForInitialization"`
+    Options devcycle.DVCOptions `json:"options"`
+}
+
+type clientResponseBody struct {
+    Message         string   `json:"message, omitempty"`
+    AsyncError      string   `json:"asyncError, omitempty"`
+    Exception       string   `json:"exception, omitempty"`
+}
+func clientHandler(w http.ResponseWriter, r *http.Request) {
```
nsalamad, line 22: I think you're missing exception handling

nsalamad: like for the case where the env key is missing we'd want to return `exception: Missing environment key....`

proxies/go/handler_client.go
```diff
@@ -0,0 +1,46 @@
...
+        return
+    } else if reqBody.ClientId == "" {
+        http.Error(w, "Invalid request: missing clientId", http.StatusBadRequest)
+        return
+    }
+
+	reqBody.Options.DisableLocalBucketing = reqBody.EnableCloudBucketing
+	client, err := devcycle.NewDVCClient(reqBody.SdkKey, &reqBody.Options, nil)
+	var res clientResponseBody
+	if err != nil {
```
nsalamad, line 36: don't know if github is being weird but there's an indentation here

proxies/go/handler_client.go
elliotCamblor: ah I think I actually missed a step entirely in the initialization process. The exception handling is going to look real strange for this endpoint because theres no way for `waitForInitialization: false` to work with the sdk right now

nsalamad: oh true, I guess there's no way to await initialization for the sdk right now

nsalamad: but we still need to wrap the whole `NewDVCClient` call in a try/catch and return `exception` in the catch

elliotCamblor: yeah, adding that now along with actually initializing local bucketing properly

proxies/go/handler_client.go
elliotCamblor: yea it looks normal on my end so im not sure lol

proxies/go/handler_client.go
```diff
@@ -0,0 +1,52 @@
...
+        return
+    }
+
+    reqBody.Options.DisableLocalBucketing = reqBody.EnableCloudBucketing
+    var lb *devcycle.DevCycleLocalBucketing
+    var res clientResponseBody
+    if reqBody.EnableCloudBucketing == false {
+        lb, err = devcycle.InitializeLocalBucketing(reqBody.SdkKey, &reqBody.Options)
+        if err != nil {
+           res.AsyncError = err.Error()
```
nsalamad, line 39: this should be `exception` cause I don't think the go proxy like the dotnet one should return an `asyncError` (see slack for details). You can also add a check now for `awaitForInitialization` and then just put a TODO in that if/else branch until we we implement async init on the sdk.

elliotCamblor: ill change the asyncError to be exception, but I dont think we know how the awaitForInitialization is going to work quite yet so I dont think theres any point in a TODO

proxies/go/handler_client.go
```diff
@@ -0,0 +1,52 @@
...
+    var res clientResponseBody
+    if reqBody.EnableCloudBucketing == false {
+        lb, err = devcycle.InitializeLocalBucketing(reqBody.SdkKey, &reqBody.Options)
+        if err != nil {
+           res.exception = err.Error()
+        }
+    }
+	client, err := devcycle.NewDVCClient(reqBody.SdkKey, &reqBody.Options, lb)
+	if err != nil {
+        res.Exception = err.Error()
```
nsalamad, line 44: the response code is 200 for the exception case

proxies/go/handler_client.go
nsalamad: kk sounds good

nsalamad (approved):

PR #140, opened 2023-01-13
Add location endpoint to go proxy

emmawillis:
this ticket isn't supposed to include parsing user and event params but it ended up being surprisingly simple so i just did it here

### DIFF
--- a/proxies/go/handler_command.go
+++ b/proxies/go/handler_command.go
@@ -2,8 +2,12 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
+	"reflect"
+	"strconv"
+	"strings"
 
 	devcycle "github.com/devcyclehq/go-server-sdk"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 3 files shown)


proxies/go/handler_command.go
```diff
@@ -4,14 +4,17 @@ import (
...
+	"strings"
 
 	devcycle "github.com/devcyclehq/go-server-sdk"
 	"github.com/gorilla/mux"
 )
 
 type Param struct {
-	Type  *string `json:"type,omitempty"`
-	Value *any    `json:"value,omitempty"`
+	Type  string `json:"type,omitempty"`
```
ajwootto, line 16: did this need to be changed? Using pointers here makes the "empty" state a nil value which I think is easier to check for

proxies/go/handler_command.go
```diff
@@ -22,11 +25,31 @@ type CommandBody struct {
 	Params  []Param            `json:"params"`
 }
 
+type LocationResponse struct {
+	EntityType string      `json:"entityType"`
+	Data       interface{} `json:"data"`
```
ajwootto, line 30: we should be consistent about using `any` vs `interface{}` (pretty sure they're the same thing)

personally I prefer `any` because it's more clear what it means (especially to all us Typescript folks)

proxies/go/handler_command.go
```diff
@@ -35,5 +58,62 @@ func commandHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	w.Header().Set("Location", locationHeader)
+	json.NewEncoder(w).Encode(response)
+
+	// TODO handle exception and asyncError
+	// TODO test with local bucketing
+}
+
+func getEntityFromLocation(location string, id string) any {
+	if location != "" {
```
ajwootto, line 99: seems like this is somewhat duplicating the logic above that's determining whether its a user request

proxies/go/handler_command.go
emmawillis: good point, i think i accidentally changed it back fixing merge conflicts - fixed!

proxies/go/handler_command.go
```diff
@@ -22,11 +25,31 @@ type CommandBody struct {
...
+
 func commandHandler(w http.ResponseWriter, r *http.Request) {
 	var body CommandBody
 	vars := mux.Vars(r)
 
-	fmt.Printf("Location: %v\n", vars["location"])
+	var location = vars["location"]
+	var id = vars["id"]
+	var locationIsClient = false
+	if location != "" {
```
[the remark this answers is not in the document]
emmawillis: i couldn't get the path parsing to work without having 2 separate handlers like
```
	r.HandleFunc("/command/{location}/{id}", commandHandler).Methods("POST")
	r.HandleFunc("/client/{id}", commandHandler).Methods("POST")

```
so it ended up being simpler to just check if location was matched or only id

proxies/go/handler_command.go
emmawillis: good point, i added locationIsClient as a param

proxies/go/handler_command.go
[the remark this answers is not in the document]
ajwootto: maybe it would be cleaner then to have two separate handler functions that both call a shared method which takes some param like "type"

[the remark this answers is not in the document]
emmawillis: sure, that makes sense - just added a clientCommandHandler and locationCommandHandler

ajwootto (approved):

PR #153, opened 2023-01-19
Add BucketingAPIURI override and parse empty map results as {} instead of null

emmawillis:
these 2 changes get allFeatures cloud passing completely and allVariables cloud mostly passing except for the 400-level exceptions aren't thrown (making a ticket for that now)

### DIFF
--- a/proxies/go/handler_command.go
+++ b/proxies/go/handler_command.go
@@ -184,15 +184,12 @@ func callMethodOnEntity(
 	}
 
 	result := method.Call(params)
-
-	entityType := method.Type().Out(0).Name()
-	if entityType != "Variable" {
-		entityType = "Object"
-	}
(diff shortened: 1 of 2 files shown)


proxies/go/handler_command.go
```diff
@@ -203,3 +200,35 @@ func callMethodOnEntity(
...
+		if r := recover(); r != nil {
+			handleError(r, err)
+		}
+	}()
+
+	var parsedResult any
+	parsedResult = result[0]
+
+	// map types need to be explicitly cast to map[string]<type> for json parsing
+	if entityType == reflect.TypeOf(make(map[string]devcycle.Variable)) {
```
ajwootto, line 215: this may be affected by Jason's changes that added the ReadOnlyVariable type

nsalamad (approved):

PR #188, opened 2023-02-17
DVC-6222 fix initialize SDK tests with new expectErrorToContain()

jonathannorris:
As some SDKs will log either: 
- `Missing SDK key! Call build with a valid SDK key`
- `Missing SDK key! Call initialize with a valid SDK key`

### DIFF
--- a/harness/features/initialize.local.test.ts
+++ b/harness/features/initialize.local.test.ts
@@ -1,7 +1,7 @@
 import {
     forEachSDK,
     wait,
-    LocalTestClient, describeCapability, expectErrorMessageToBe
+    LocalTestClient, describeCapability, expectErrorMessageToBe, expectErrorToContain
 } from '../helpers'
 import { Capabilities } from '../types'
 import { config, shouldBucketUser } from '../mockData'
@@ -17,9 +17,9 @@ describe('Initialize Tests - Local', () => {
                 const response = await testClient.createClient(true, {}, null, true)
                 const { exception } = await response.json()
 
-                expectErrorMessageToBe(
+                expectErrorToContain(
                     exception,
-                    'Missing environment key! Call initialize with a valid environment key'
+                    ['Missing SDK key! Call ' + ' with a valid SDK key']
                 )
             })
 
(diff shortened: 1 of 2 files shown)


harness/features/initialize.local.test.ts
```diff
@@ -28,9 +28,9 @@ describe('Initialize Tests - Local', () => {
                 const response = await testClient.createClient(true, {}, 'invalid key', true)
                 const { exception } = await response.json()
 
-                expectErrorMessageToBe(
+                expectErrorToContain(
                     exception,
-                    'Invalid environment key provided. Please call initialize with a valid server environment key'
+                    ['Invalid SDK key provided. Please call ' + ' with a valid server SDK key']
```
ajwootto, line 33: is this `+` supposed to be a comma?

jonathannorris: yes

harness/helpers.ts
```diff
@@ -442,6 +442,15 @@ export class CloudTestClient extends BaseTestClient {
     }
 }
 
+export const expectErrorToContain = (message: string, expectedStrings: string[]) => {
+    expectedStrings.forEach((expected) => {
+        if (!message.includes(expected)) {
+            console.warn(`Expected error message to contain: \n ${expected}, but got: \n ${message}`)
+        }
+        expect(message).toContain(expected)
```
ajwootto, line 450: there isnt much point in doing a warning log if you're going to actually fail on this line anyway

jonathannorris: actually looking at the code again for `expectErrorMessageToBe()` below we aren't actually enforcing the message... just doing a `expect(message).toBeTruthy()`... why is that?

image

ajwootto: because none of the SDKs throw error messages that are the same, so rather than spending the time right now to fix them all we opted to just make sure there is an error in general and then log a warning if they don't match. Eventually we want to make them match but not right now

jonathannorris: ok then I guess I can scrap most of these changes then. need to figure out what else broke this then: https://github.com/DevCycleHQ/test-harness/actions/runs/4205682656/jobs/7298129382

ajwootto: looks like something crashed the dotnet proxy server

ajwootto (approved):

PR #189, opened 2023-02-22
Test Harness Updates for Java Server SDK

kaushalkapasi:
- changes for Java SDK testing
- Currently only enabled for LOCAL mode for Java SDK

### DIFF
--- a/harness/types/sdks.ts
+++ b/harness/types/sdks.ts
@@ -1,8 +1,9 @@
 export const Sdks = {
     nodejs: 'NodeJS',
     dotnet: 'DotNet',
+    go: 'Go'
+    // TODO uncomment once Java is ready
+    // java: 'Java'
     // TODO add once python is ready
     // python: "Python"
-    // go: 'Go',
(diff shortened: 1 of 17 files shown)


suthar26 (approved):
Looks good to me, I think location endpoint and command results have been reviewed couple times and functionality wise have worked.

harness/types/sdks.ts
```diff
@@ -1,7 +1,8 @@
 export const Sdks = {
     nodejs: 'NodeJS',
     dotnet: 'DotNet',
+    go: 'Go'
     // TODO add once python is ready
```
chantalchung, line 5: update comment to say once Java is ready, too? since we're only running the Test Harness with the Java sdk in local mode right now right?

chantalchung (approved):

harness/types/sdks.ts
kaushalkapasi: yup, will add it in 😄

PR #190, opened 2023-02-22
strengthen variable event tests to check for every field

ajwootto:
- add `jest-extended` to get some extra matchers
- match on every expected field of the event object rather than doing a partial match
- this tests that an empty featureVars object is sent correctly

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -5,5 +5,6 @@ module.exports = {
         '^.+\\.tsx?$': ['ts-jest', {}]
     },
     testEnvironment: './jest-environment.ts',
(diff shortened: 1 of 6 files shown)


jest.config.ts
```diff
@@ -5,5 +5,6 @@ module.exports = {
         '^.+\\.tsx?$': ['ts-jest', {}]
     },
     testEnvironment: './jest-environment.ts',
+    "setupFilesAfterEnv": ["./jest-setup.js"],
```
jonathannorris, line 8:
```suggestion
    setupFilesAfterEnv: ["./jest-setup.js"],
```

jonathannorris (approved):

PR #191, opened 2023-02-22
DVC-5973 [fix] send body when intercepting request

chantalchung:
- java sdk expects a body, throws exception without
- verified tests continue to pass on all sdks


kaushalkapasi:
I don't think this is the right change... it will cause other SDKS to fail when running against the test-harness?

PR #201, opened 2023-02-27
add /client endpoint for Ruby proxy

nsalamad:


proxies/ruby/app.rb
```diff
@@ -15,3 +16,52 @@
...
+    data.values_at('clientId', 'sdkKey', 'waitForInitialization', 'options')
+
+  if client_id.nil?
+    status 400
+    return { message: 'Invalid request: missing clientId' }.to_json
+  end
+
+  begin
+    options = DevCycle::DVCOptions.new(
+      config_cdn_uri: options.fetch('configCDNURI', 'https://config-cdn.devcycle.com'),
```
JamieSinn, line 39: Is this how the defaulting works?

JamieSinn (approved):

proxies/ruby/app.rb
nsalamad: Yup. I needed to do that, otherwise we'd be passing in nils and that would cause the sdk to throw

JamieSinn: TIL

PR #206, opened 2023-02-28
Add action to be triggered from other SDK repos.

JamieSinn:


action.yml
```diff
@@ -0,0 +1,41 @@
+name: 'DevCycle Test Harness'
+description: 'Run test harness tests against given SDKs'
+inputs:
+  sdk:
```
ajwootto, line 4: technically this is meant to potentially be a list of SDKs to test, which would make this `sdks`

also might as well just use the same name as the env var to prevent confusion, `sdks-to-test`

JamieSinn: fixed

ajwootto (approved):

PR #213, opened 2023-03-01
add /command endpoint for Ruby proxy

nsalamad:


proxies/ruby/app.rb
```diff
@@ -19,49 +23,89 @@
...
     status 400
     return { message: 'Invalid request: missing clientId' }.to_json
   end
 
   begin
-    options = DevCycle::DVCOptions.new(
-      config_cdn_uri: options.fetch('configCDNURI', 'https://config-cdn.devcycle.com'),
-      events_api_uri: options.fetch('eventsAPIURI', 'https://events.devcycle.com'),
+    dvc_options = DevCycle::DVCOptions.new(
+      config_cdn_uri: options.fetch('configCDNURI', ''),
```
JamieSinn, line 43: why are we defaulting to empty string now?

proxies/ruby/helpers.rb
```diff
@@ -0,0 +1,58 @@
...
+  case type
+  when 'user'
+    req_body_user = body['user']
+    user_data = DevCycle::UserData.new
+    user_data.user_id = req_body_user['user_id'] if req_body_user['user_id']
+    user_data.email = req_body_user['email'] if req_body_user['email']
+    user_data.name = req_body_user['name'] if req_body_user['name']
+    user_data.language = req_body_user['language'] if req_body_user['language']
+    user_data.country = req_body_user['country'] if req_body_user['country']
+    user_data.app_version = req_body_user['appVersion'] if req_body_user['appVersion']
```
JamieSinn, line 41: are these not camel cased?
https://github.com/DevCycleHQ/ruby-server-sdk/blob/main/lib/devcycle-ruby-server-sdk/models/user_data.rb#L39

proxies/ruby/app.rb
nsalamad: there's no point in defaulting to our actual config CDN and events api. We're not using real sdk keys for the test harness so we'd be making requests just to have them fail

proxies/ruby/helpers.rb
nsalamad: yep, you're right. Will fix

proxies/ruby/app.rb
```diff
@@ -19,49 +23,89 @@
...
-
     { message: 'success' }.to_json
   rescue StandardError => e
     status 200
-    { error: e.message }.to_json
+    { exception: e.message }.to_json
+  end
+end
+
+post '/*' do
```
JamieSinn, line 61: is this the correct route?

JamieSinn: yeah - the regex would help a lot

proxies/ruby/app.rb
ajwootto: we also really do not want this thing making real requests by default

proxies/ruby/app.rb
```diff
@@ -19,49 +22,89 @@
...
     status 201
     headers 'Location' => "/client/#{client_id}"
 
-    return { asyncError: async_error }.to_json if async_error
-
     { message: 'success' }.to_json
   rescue StandardError => e
     status 200
-    { error: e.message }.to_json
+    { exception: e.message }.to_json
```
ajwootto, line 56: do we always send initialization errors as "exception"? is that why this change?

nsalamad: for nodejs we have code that sends `exception` or `asyncError` based on where the error occurs but the `clientInitialized` never actually throws so we decided for Go and Dotnet (Ruby now too) to never report `asyncErorrors`

JamieSinn (approved):

PR #232, opened 2023-03-20
add test for receiving new config during test and seeing new variable

ajwootto:
- add new multi threading test that simulates receiving a new config and tests that the variable call correctly returns the new value
- also fixes multithreaded access to commandResults in Go proxy by putting a mutex around the operation that stores the results

### DIFF
--- a/proxies/go/handler_command.go
+++ b/proxies/go/handler_command.go
@@ -62,8 +62,6 @@ func locationCommandHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func commandHandler(locationIsClient bool, w http.ResponseWriter, r *http.Request) {
-	commandMutex.Lock()
-	defer commandMutex.Unlock()
 	var body CommandBody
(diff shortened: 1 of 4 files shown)


proxies/go/handler_command.go
```diff
@@ -228,6 +226,9 @@ func callMethodOnEntity(
 	locationId := strconv.Itoa(len(datastore.commandResults[command]))
 	locationHeader := "command/" + command + "/" + locationId
 
+	commandMutex.Lock()
+	defer commandMutex.Unlock()
```
rob-odwyer, line 230: I think you still need to lock around the reads as well.

Since maps are pointer types, might be able to make a `getMap` method that does the locking and returns a reference to it, then use that everywhere instead?

rob-odwyer: Ignore that idea - just remembered that you won't be able to defer the unlock that way.

I think we still need at least a `sync.RWMutex` for every access to the map though.

PR #246, opened 2023-05-01
DVC-7110 Added test harness proxy for OpenFeature NodeJS Provider

jonathannorris:


harness/features/variable.local.test.ts
```diff
@@ -136,39 +137,46 @@ describe('Variable Tests - Local', () => {
...
-                        // Expect that the test returns a defaulted variable
-                        expectDefaultValue(key, variable, wrongTypeDefault,
-                            wrongTypeDefault === '1' ? VariableType.string : VariableType.number)
-
-                        if (!hasCapability(name, Capabilities.events)) {
-                            return
-                        }
+                    // Skip test for OpenFeature NodeJS SDK as it throws the error in the OF Provider
+                    // and doesn't log events which doesn't seem like something we can easily disable checking for.
+                    if (name !== 'OF-NodeJS') {
```
jonathannorris, line 142: tried this test by disabling `interceptEvents()` and exiting at `hasCapability` below, but got error where it was expecting a `/events/batch` call still.

proxies/openfeature-nodejs/handlers/client.ts
```diff
@@ -0,0 +1,69 @@
...
+                } catch (e) {
+                    asyncError = e
+                }
+            }
+
+        } else {
+            dvcClient = initialize(sdkKey, { ...options, enableCloudBucketing: true })
+        }
+
+        // @ts-ignore
```
jonathannorris, line 52: This is only because of the local yarn link to the openfeature provider. will remove.

proxies/openfeature-nodejs/handlers/location.ts
```diff
@@ -0,0 +1,258 @@
...
+        )
+    } else {
+        throw new Error('Invalid default value type')
+    }
+}
+
+/**
+ * Fake DVCVariable Class so that the variable type reporting works correctly
+  */
+class DVCVariable implements DVCVariableInterface {
```
jonathannorris, line 184: this is a bit gross but it works. this is because the NodeJS SDK returns a `DVCVariable` class, but we only expose the interface publically.

proxies/openfeature-nodejs/handlers/location.ts
```diff
@@ -0,0 +1,258 @@
...
+}
+
+const invokeCommand = (
+    entity: DataStoreClient | DVCUser | DVCVariable | any,
+    command: string,
+    params: ParsedParams,
+) => {
+    console.log(`invoking command ${command} on ${typeof entity} with params ${JSON.stringify(params)}`)
+    const dataStoreClient = entity as DataStoreClient
+    if (command === 'variable') {
```
jonathannorris, line 231: This is a bit different too, basically only calls to `variable` are passed through the OpenFeature client, all others are sent to the DVCClient

harness/features/variable.local.test.ts
rob-odwyer: I think you might need to remove `Capabilities.events` from the corresponding record in capabilities.ts?

chris-hoefgen (approved):

PR #249, opened 2023-05-11
PHP Test Harness - Proxy mode

JamieSinn:


proxies/php/nginx/nginx.conf
```diff
@@ -0,0 +1,35 @@
...
+        proxy_pass http://127.0.0.1:8000$request_uri;
+    }
+
+    location ~ /close/.* {
+            proxy_pass http://127.0.0.1:8000$request_uri;
+        }
+    # Match the /<clientId>/v1/.* pattern to route to the correct socket proxy
+    location ~ ^/(.*)(/v1/.*)$ {
+        proxy_pass http://unix:/tmp/$1.sock:$2;
+    }
```
rob-odwyer, line 18: I think the above block is unused? The PHP SDK seems to use the Unix socket directly, and removing this had no effect on the tests that I could see.

proxies/php/sidecar-manager/go.mod
```diff
@@ -0,0 +1,40 @@
+module github.com/devcyclehq/test-harness/php/sidecar-manager
+
+go 1.20
+
+require (
+	github.com/devcyclehq/go-server-sdk/v2 v2.9.7-0.20230503141249-05a8cad604f1
```
rob-odwyer, line 6:
```suggestion
	github.com/devcyclehq/go-server-sdk/v2 v2.10.2
	github.com/devcyclehq/local-bucketing-proxy v0.0.4
```

proxies/php/php-proxy/index.php
```diff
@@ -0,0 +1,197 @@
...
+{
+    echo json_encode(["name" => "PHP", "version" => "1.0.0", "capabilities" => ["LocalBucketing", "CloudBucketing", "Events"]]);
+    exit(200);
+}
+
+function buildClient(string $clientId)
+{
+    $config = DevCycle\Configuration::getDefaultConfiguration()
+        ->setApiKey('Authorization', getenv("LOCAL_MODE") == "" ? "dvc_server_" . $clientId : "dvc_server_test-harness-config")
+        ->setHost(getenv("LOCAL_MODE") == "" ? "http:/" . $clientId . "v1" : "http://localhost:8080")
```
rob-odwyer, line 141: Is setting the host here used at all, given the unix domain socket is set next?
It looks to me like it's constructing an invalid URL (starting with `http:/`)

rob-odwyer (changes requested):

proxies/php/nginx/nginx.conf
JamieSinn: This is to allow you to take a test locally and run it against the container if you breakpoint the test harness.

It's a thing I was using during development, but may be useful otherwise?

harness/helpers.ts
```diff
@@ -29,6 +29,7 @@ export const getConnectionStringForProxy = (proxy: string) => {
 
     const host = global[`__TESTCONTAINERS_${proxy.toUpperCase()}_IP__`]
     const port = global[`__TESTCONTAINERS_${proxy.toUpperCase()}_PORT_3000__`]
+    //const port = 3000
```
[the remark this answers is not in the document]
JamieSinn: Yep, should be removed.

proxies/php/php-proxy/index.php
JamieSinn: That's how curl/php access Unix sockets. I can find docs, but that's setting the protocol, and route path

proxies/php/startup.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+/sidecar-manager &
```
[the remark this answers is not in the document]
JamieSinn: Definitely could, but it failing in supervisord would cause a similar failure visibility would it not?

proxies/php/php-proxy/index.php
rob-odwyer: It seems to still work if I correct it to `http://`, but I'll take your word for it.

proxies/php/startup.sh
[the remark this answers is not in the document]
rob-odwyer: Yeah, good point - we haven't really solved the overall problem of seeing what's going on inside the proxies when running the test harness in docker.
If there's no reason to run it separately here, I'd still prefer to run this through supervisord so it's easy to find, but it's not critical.

rob-odwyer (approved):

proxies/php/php-proxy/index.php
JamieSinn: That's because you're technically setting the protocol to http:/ and then the path as /, which is valid

PR #252, opened 2023-05-31
DVC-7132 Java Server SDK support

chris-hoefgen:
updated the proxy, build process and docker container to bring it up to date and work with the Java Server SDK


rob-odwyer (approved):
The timeout issue seems to happen on main as well, so this should be safe to merge since it passes in isolation.

PR #257, opened 2023-06-13
DVC-7569 enable python sdk

chris-hoefgen:
- Finished missing functionality in the python proxy
- got all the tests working for the cloud SDK
- fixed the Dockerfile to load versions and SHA code

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -82,8 +82,7 @@ const projects = [
         globals: {
             JEST_PROJECT_SDK_TO_TEST: 'Python',
             LOCAL_HOST_BINDING: '0.0.0.0',
-        },
-        runner: 'jest-serial-runner'
+        }
     },
 ]
 const SDKs = getSDKs().map((sdkName) => sdkName.toLowerCase())
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -75,15 +75,16 @@ const projects = [
             LOCAL_HOST_BINDING: '0.0.0.0'
         },
         runner: 'jest-serial-runner'
-    }
-    // TODO uncomment once Python is ready
-    // {
-    //     ...commonConfig,
-    //     displayName: 'Python',
-    //     globals: {
-    //         JEST_PROJECT_SDK_TO_TEST: 'Python',
-    //     }
-    // },
+    },
+    {
+        ...commonConfig,
+        displayName: 'Python',
+        globals: {
+            JEST_PROJECT_SDK_TO_TEST: 'Python',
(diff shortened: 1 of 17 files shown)


jest.config.ts
```diff
@@ -75,15 +75,16 @@ const projects = [
...
-    // },
+    },
+    {
+        ...commonConfig,
+        displayName: 'Python',
+        globals: {
+            JEST_PROJECT_SDK_TO_TEST: 'Python',
+            LOCAL_HOST_BINDING: '0.0.0.0',
+        },
+        runner: 'jest-serial-runner'
```
rob-odwyer, line 86: Do we need the serial runner for Python? It seems to work for me locally without, and I'm not aware of any concurrency issues (yet).

rob-odwyer: OK, it seems to actually run *faster* with the serial runner enabled. So maybe we should leave it in place.

chris-hoefgen: good point.  you are right in that it probably doesn't.  I think i put it there to help with dev/debugging

chris-hoefgen: i'll remove that

chris-hoefgen: lol.  i'll put it back

proxies/python/install.sh
```diff
@@ -1,6 +1,17 @@
...
+#!/bin/sh
+
+if [[ -n "$SDK_GITHUB_SHA" ]]; then
+    echo "Checking out python SDK at $SDK_GITHUB_SHA"
+    cd ..
+    git clone https://github.com/DevCycleHQ/python-server-sdk.git
+    cd python-server-sdk
+    git checkout $SDK_GITHUB_SHA
+    cd ..
+    pip3 install -e ../python-server-sdk/
```
rob-odwyer, line 10: These paths don't seem quite right to me, but maybe I'm missing something.
We go up one directory, then clone the SDK repo and cd into it, then back up to the parent directory again, but then reference the `python-server-sdk` from the parent directory?

chris-hoefgen: yeah i see what you mean.  the second `cd ..` really doesn't make sense. Like if anything shouldn't it be trying to go back down into the app/ folder where the proxy code is running?

chris-hoefgen: I think it is supposed to be `cd ../app` to get back into the proxy directory

chris-hoefgen: ah, i think the pip install is global since this is run as root so the directory change is likely unnecessary.  See my latest change.

PR #260, opened 2023-06-26
Output full logs for docker build in CI

ajwootto:
- adds an output of full logs from docker builds during CI run


chris-hoefgen (approved):
LGTM - seems to match the docs

PR #267, opened 2023-07-18
fix: [DVC-8210] update JS SDKs to new SDK interfaces

jonathannorris:


proxies/nodejs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "dependencies": {
-    "@devcycle/nodejs-server-sdk": "latest",
+    "@devcycle/nodejs-server-sdk": "*",
```
chris-hoefgen, line 3: Just curious, what is the purpose of this change?

chris-hoefgen (approved):
LGTM.

proxies/nodejs/package.json
jonathannorris: I didn't notice that the package was getting updated when it was set to `latest`, which should be the same thing as setting it to `*`.

PR #352, opened 2024-01-09
ensure bash scripts exit on error

ajwootto:
the install.sh script wasn't aborting early when a command failed, it was just proceeding, causing us to miss installation issues


proxies/openfeature-nodejs/Dockerfile
```diff
@@ -15,5 +15,6 @@ RUN apk update && apk upgrade && \
 COPY . /app
 WORKDIR /app
 RUN yarn
+RUN echo "Invalidate"
```
jonathannorris: is this still needed?

jonathannorris (approved):

jonathannorris (approved):

PR #358, opened 2024-01-10
fix: openfeature test harness mising .yarnrc.yml file, and node not running install.sh

jonathannorris:


jonathannorris:
Current dependencies on/for this PR:
* `main`
  * **PR #358** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/test-harness/358?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/DevCycleHQ/test-harness/358?utm_source=stack-comment).

proxies/openfeature-nodejs/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "dependencies": {
     "@devcycle/openfeature-nodejs-provider": "*",
-    "@openfeature/core": "^0.0.22",
-    "@openfeature/server-sdk": "^1.9.0",
+    "@openfeature/core": "*",
```
ajwootto, line 4: putting * here doesn't change anything, it will just be locked to whatever version it first resolves to

ajwootto (approved):

proxies/openfeature-nodejs/package.json
jonathannorris: that's fine, but the intention with these is to use the latest, which I think setting as `*` implies.

ajwootto: it doesn't really, I think it falsely gives the impression that it will always use latest when it actually won't

PR #410, opened 2024-04-22
add test for config etag caching and reporting

ajwootto:
- add new test for config etag changing and verify bucketing result changes
- also check the etag being reported in event metadata
- update the mock server to properly forward headers send by nock response


jonathannorris (approved):

suthar26 (approved):
this works with the local version of the sdk

PR #411, opened 2024-04-22
feat: add a test for when config is changed and etag reporting is correct

suthar26:


harness/features/variable.local.test.ts
```diff
@@ -70,15 +69,96 @@ describe('Variable Tests - Local', () => {
             }
         }
 
+        describe('etag test client', () => {
+
+            it('should respond with the correct etag', async () => {
```
ajwootto, line 74: is this testing anything different than what's already here?

suthar26:
closing as this is tested in the initalize tests

PR #420, opened 2024-04-29
Disable SSE during test harness until we write functionality for it.

JamieSinn:
The delayed startup in the case of SSE failing to connect causes tests to fail due to defaulting. Also updated the deprecated names


jonathannorris (approved):

proxies/go/handler_client.go
```diff
@@ -59,10 +59,12 @@ func clientHandler(w http.ResponseWriter, r *http.Request) {
 		EventFlushIntervalMS:    time.Duration(reqBody.Options.EventFlushIntervalMS * 1000000),
 		EnableCloudBucketing:    reqBody.EnableCloudBucketing,
 		OnInitializedChannel:    onInitializedChannel,
+		// TODO: Properly implement SSE tests. Currently the init delay on config pull breaks tests.
+		DisableServerSentEvents: true,
```
ajwootto, line 63: this field should be named `disableRealtimeUpdates` to match the other SDKs

proxies/go/go.mod
```diff
@@ -3,23 +3,30 @@ module github.com/devcyclehq/test-harness/proxies/go
 go 1.19
 
 require (
-	github.com/devcyclehq/go-server-sdk/v2 v2.9.5
+	github.com/devcyclehq/go-server-sdk/v2 v2.13.1-0.20240429135854-e299e6ccb33a
```
ajwootto, line 6: whats with the stuff on the end of the version tag?

JamieSinn: this is the direct hash reference to the SSE branch to allow  using that property.

proxies/go/handler_client.go
JamieSinn: We aren't calling it realtime updates for server SDKs as far as i know - it's being called SSE as far as I know? @jonathannorris 

Realtime updates is the client sdk concept

jonathannorris: Yeah, it's the same thing, it should be named the same.

JamieSinn: kk - i'll update the SDK then

jonathannorris: https://docs.devcycle.com/sdk/client-side-sdks/javascript/javascript-gettingstarted#initialization-options

ajwootto: SSE is the name of a technology which powers the "realtime updates" feature, SSE by itself would be a meaningless term to most people I think

proxies/go/go.mod
ajwootto: we can't commit that can we?

JamieSinn: We can - it doesn't actually matter because the install script overrides it anyway.
This is just to ensure that the local build doesn't fail

ajwootto: when running locally shouldnt you use the local linking version of the dependency that's commented out below?

JamieSinn: That's an option - but the way I was debugging is to ref to the existing git ref to prevent needing to run both at once and manually attach debuggers to both - this allows you to just run one debug session in IntelliJ/WebStorm - given you know that the test harness is the problem. 

It's just two approaches - but without this change the local code will fail and not build until we merge the SSE PR in Go

ajwootto: ok just remember to update it after

PR #426, opened 2024-05-02
add bootstrapping tests

ajwootto:
Add tests for Nodejs Bootstrapping mode

Currently requires the proxy to also install the js SDK due to the dynamic import being used for bootstrapping the clientside user, which maybe isn't ideal


proxies/nodejs/package.json
```diff
@@ -12,5 +13,10 @@
     "devDependencies": {
         "typescript": "^4.9.3"
     },
-    "packageManager": "yarn@4.1.1"
+    "packageManager": "yarn@4.1.1",
+    "resolutions": {
+        "@devcycle/bucketing-assembly-script": "portal:../../../js-sdks/lib/shared/bucketing-assembly-script",
+        "@devcycle/js-client-sdk": "portal:../../../js-sdks/dist/sdk/js",
+        "@devcycle/types": "portal:../../../js-sdks/dist/lib/shared/types"
```
jonathannorris, line 20: Don't you need to remove these?

ajwootto: ya

jonathannorris (approved):

PR #432, opened 2024-05-08
feat: SSE Prep/Test Cleanup

JamieSinn:
This enables the test harness to generate SSE specific configs if needed and cleans up the repeated config paths ad infinitum to DRY - and make it much cleaner.

Also sets up a proxy options type to better keep track of all the options that the proxies should/could implement.


harness/helpers/helpers.ts
```diff
@@ -437,11 +441,21 @@ export class LocalTestClient extends BaseTestClient {
...
         return result
     }
+
+    getValidConfigPath() {
+        return `/client/${this.clientId}/config/v1/server/${this.sdkKey}.json`
+    }
+
+    getValidConfig() {
+        return this.options.enableRealtimeUpdates
+            ? config_sse(getMockServerUrl(), `/client/${this.clientId}/sse`)
```
ajwootto, line 451: why not just always send the correct SSE paths in the config?

JamieSinn: Dunno - wanted this to be as minimal breaking as possible and that's what felt like an easy option/safe?

harness/features/track.local.test.ts
```diff
@@ -36,10 +35,8 @@ describe('Track Tests - Local', () => {
             url = getConnectionStringForProxy(sdkName)
 
             scope
-                .get(
-                    `/client/${client.clientId}/config/v1/server/${client.sdkKey}.json`,
-                )
-                .reply(200, config)
+                .get(client.getValidConfigPath())
+                .reply(200, client.getValidConfig)
```
ajwootto, line 39: passing the function like this might have issues with binding to the `this` context of the client instance, like if this function is using the `this` keyword in it then it wont work

harness/helpers/helpers.ts
ajwootto: the configs every SDK is pulling down are gonna have these SSE properties in them, so might as well make it consistent

harness/features/track.local.test.ts
JamieSinn: wait what? can you explain more?

harness/helpers/helpers.ts
JamieSinn: kk - will fix

harness/features/track.local.test.ts
ajwootto: basically in JS  a function is bound to the context its called from, so if you do
```client.someMethod()```
then within that method `this` refers to the client
but if you do something like 
```
const method = client.someMethod
method()
```
then when invoked `this` will be undefined because its not being called within the same context
to fix it you can either do:
```
const method = client.someMethod.bind(client)
method()
```

which makes it so that the method is always bound to "client"

or you can use an arrow function like

```
const method = () => client.someMethod()
method()
```
to basically "wrap" it while preserving context

you can also write the method as an arrow function on the class itself

```
class Client {
 method = () => {
}
}
```
which then makes it always use `this` of the client

File: harness/features/allFeatures.local.test.ts [hunk 3]
```diff
@@ -34,10 +34,8 @@ describe('allFeatures Tests - Local', () => {
             beforeEach(async () => {
                 testClient = new LocalTestClient(sdkName)
                 scope
-                    .get(
-                        `/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`,
-                    )
-                    .reply(200, config)
+                    .get(testClient.getValidConfigPath())
+                    .reply(200, testClient.getValidConfig())
```
ajwootto, line 38: you could also just make this whole scope thing a single function on the client like `respondWithConfig(scope)`

harness/features/initialize.local.test.ts
```diff
@@ -310,11 +274,9 @@ describe('Initialize Tests - Local', () => {
         it('uses the new config when etag changes, and flushes existing events', async () => {
             const testClient = new LocalTestClient(sdkName)
             scope
-                .get(
-                    `/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`,
-                )
+                .get(testClient.getValidConfigPath())
                 .times(1)
-                .reply(200, config, {
+                .reply(200, testClient.getValidConfig, {
```
ajwootto, line 279: call it here

harness/features/multithreading.local.test.ts
```diff
@@ -38,10 +37,8 @@ describe('Multithreading Tests', () => {
                 testClient = new LocalTestClient(sdkName)
 
                 scope
-                    .get(
-                        `/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`,
-                    )
-                    .reply(200, config, {
+                    .get(testClient.getValidConfigPath())
+                    .reply(200, testClient.getValidConfig, {
```
ajwootto, line 41: call this

harness/features/allFeatures.local.test.ts
JamieSinn: I was debating that - how would i pass through the scope/whatnot?

harness/features/initialize.local.test.ts
JamieSinn: missed one

harness/features/allFeatures.local.test.ts
ajwootto: just give it as an argument

JamieSinn: isn't js pass by value not by ref?

ajwootto: or there may be a way to retrieve it from somewhere automatically since the scope is set up automatically by the test environment

ajwootto: > isn't js pass by value not by ref?
nope

JamieSinn: yes - but we also override the flow of responding successfully in a few cases - and wanted to keep that functionality

ajwootto: yeah  you can still construct the responder manually this would just be a slight convenience for the most common case

harness/features/variable.local.test.ts
```diff
@@ -84,10 +83,8 @@ describe('Variable Tests - Local', () => {
                 // conflicting. This one is used to mock the config that the local client is going to use
                 // locally in all of its methods.
                 scope
-                    .get(
-                        `/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`,
-                    )
-                    .reply(200, config, {
+                    .get(testClient.getValidConfigPath())
+                    .reply(200, testClient.getValidConfig, {
```
ajwootto, line 87: call this

harness/features/allFeatures.local.test.ts
JamieSinn: yeah - gonna do a client function to make it cleaner

harness/features/allFeatures.local.test.ts
```diff
@@ -32,34 +32,33 @@ describe('allFeatures Tests - Local', () => {
...
+                .reply(200, testClient.getValidConfig)
             await testClient.createClient(true, {
                 configPollingIntervalMS: 60000,
             })
         })
 
-        it('should return all features for user without custom data', async () => {
-            const featuresResponse = await testClient.callAllFeatures({
-                user_id: 'user3',
+            it('should return all features for user without custom data', async () => {
```
jonathannorris, line 51: why is this tabbed in?

JamieSinn: I have no idea, prettier does what prettier wants

harness/features/bootstrapping.test.ts
```diff
@@ -11,12 +10,7 @@ describe('Bootstrapping Tests', () => {
 
         describe('not enabled client', () => {
             beforeEach(async () => {
-                testClient = new LocalTestClient(sdkName)
-                scope
```
jonathannorris, line 15: why is this removed?

harness/features/bootstrapping.test.ts
```diff
@@ -43,12 +37,7 @@ describe('Bootstrapping Tests', () => {
 
         describe('initialized client', () => {
             beforeEach(async () => {
-                testClient = new LocalTestClient(sdkName)
-                scope
```
jonathannorris, line 47: why is this removed?

harness/features/initialize.local.test.ts
```diff
@@ -243,16 +239,11 @@ describe('Initialize Tests - Local', () => {
 
     it('uses the same config if the refetch fails, after retrying once', async () => {
         const testClient = new LocalTestClient(sdkName)
+        addClientSDKConfigRequestMock(testClient)
 
         scope
```
jonathannorris, line 244: removing the first 200 response here kinda breaks the point of this test.

harness/features/bootstrapping.test.ts
JamieSinn: I'm not sure, but the test didn't break? Might be from my rebase?

harness/features/initialize.local.test.ts
```diff
@@ -356,10 +339,10 @@ describe('Initialize Tests - Local', () => {
         const testClient = new LocalTestClient(sdkName)
         scope
             .get(
-                `/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`,
+                testClient.getValidConfigPath(),
             )
             .times(1)
-            .reply(200, config, {
+            .reply(200, testClient.getValidConfig, {
```
jonathannorris, line 345:
```suggestion
            .reply(200, testClient.getValidConfig(), {
```

harness/features/bootstrapping.test.ts
JamieSinn: Oh, it was from the rebase, it is lower down

harness/features/initialize.local.test.ts
jonathannorris: would probably just be easier to pull out config at the start of this test into a `config` const rather than keep calling `testClient.getValidConfig()`

harness/features/variable.local.test.ts
```diff
@@ -66,22 +66,20 @@ describe('Variable Tests - Local', () => {
...
-            scope
-                .get(
-                    `/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`,
-                )
-                .reply(200, config, {
-                    ETag: 'local-var-etag',
-                    'Cf-Ray': 'local-ray-id',
-                    'Last-Modified': lastModifiedDate.toUTCString(),
-                })
+        describe('initialized client', () => {
```
jonathannorris, line 69: why is this tabbed in more?

harness/helpers/helpers.ts
```diff
@@ -273,14 +275,22 @@ class BaseTestClient {
...
     private shouldFailInit = false
+    private options: ProxyClientOptions
 
+    constructor(sdkName: string, scope?: Scope) {
+        super(sdkName)
+        if (scope) {
+            this.setupMockConfig(scope)
+        }
+    }
     async createClient(
```
jonathannorris, line 286:
```suggestion
    }
    
    async createClient(
```

harness/features/variable.local.test.ts
JamieSinn: can retry prettier - but I have no idea

harness/features/initialize.local.test.ts
JamieSinn: I kinda wanted to just make it part of the constructor and have a param to override and use a custom pull

harness/features/bootstrapping.test.ts
JamieSinn: The scope param was passed in - so it sets it up in the constructor

harness/features/bootstrapping.test.ts
JamieSinn: same as above - it uses the scope param

.github/dependabot.yml
```diff
@@ -1,71 +0,0 @@
-version: 2
-updates:
```
jonathannorris, line 2: why did you remove all these?

JamieSinn: Dependabot will still give us security updates- this prevents the nagging PRs

harness/features/allFeatures.local.test.ts
```diff
@@ -31,13 +31,7 @@ describe('allFeatures Tests - Local', () => {
         let testClient: LocalTestClient
 
         beforeEach(async () => {
-            testClient = new LocalTestClient(sdkName)
-            scope
```
jonathannorris, line 35: why is this not setting up the config mock anymore?

JamieSinn: it's set in the client constructor

harness/features/bootstrapping.test.ts
jonathannorris: what constructor? like this this is the code on main right now, is it not setting up/testing for the config mock anymore?

harness/helpers/helpers.ts
```diff
@@ -273,14 +275,22 @@ class BaseTestClient {
 
 export class LocalTestClient extends BaseTestClient {
     private shouldFailInit = false
+    private options: ProxyClientOptions
 
+    constructor(sdkName: string, scope?: Scope) {
+        super(sdkName)
+        if (scope) {
+            this.setupMockConfig(scope)
+        }
```
JamieSinn, line 284: @jonathannorris here is the setup for the scope

harness/helpers/helpers.ts
```diff
@@ -415,11 +426,25 @@ export class LocalTestClient extends BaseTestClient {
...
+
+    getValidConfigPath() {
+        return `/client/${this.clientId}/config/v1/server/${this.sdkKey}.json`
+    }
+
+    setupMockConfig(scope: Scope) {
+        scope
+            .get(this.getValidConfigPath())
+            .reply(200, this.getValidConfig())
+    }
```
JamieSinn: @jonathannorris  and this is where it sets up the mock config

harness/features/initialize.local.test.ts
jonathannorris: also you are still not calling the function here, `()` are missing.

JamieSinn: ah - just didn't push it locally

harness/features/variable.local.test.ts
```diff
@@ -74,10 +73,8 @@ describe('Variable Tests - Local', () => {
             // conflicting. This one is used to mock the config that the local client is going to use
             // locally in all of its methods.
             scope
-                .get(
-                    `/client/${testClient.clientId}/config/v1/server/${testClient.sdkKey}.json`,
-                )
-                .reply(200, config, {
+                .get(testClient.getValidConfigPath())
+                .reply(200, testClient.getValidConfig, {
```
jonathannorris, line 77: Don't you need to call the `getValidConfig` method?

JamieSinn: yep - fixed in the most recent push

jonathannorris (approved):

PR #433, opened 2024-05-10
feat: add support for new sdkConfig event from NodeJS

jonathannorris:


jonathannorris:
* **#433** <a href="https://app.graphite.dev/github/pr/DevCycleHQ/test-harness/433?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @jonathannorris and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

harness/features/variable.local.test.ts
```diff
@@ -550,6 +550,7 @@ describe('Variable Tests - Local', () => {
                             etag: null,
                             rayId: null,
                             lastModified: null,
+                            skipSDKConfigEvent: true,
```
suthar26, line 553: just to confirm, this is because of the client not being initialized?

jonathannorris: yea

harness/features/allVariables.local.test.ts
```diff
@@ -53,12 +47,30 @@ describe('allVariables Tests - Local', () => {
...
 
         it('should return a variable map for a bucketed user', async () => {
+            const client = new LocalTestClient(sdkName)
+            scope
+                .get(
+                    `/client/${client.clientId}/config/v1/server/${client.sdkKey}.json`,
+                )
+                .reply(200, config)
+
+            if (hasCapability(sdkName, Capabilities.sdkConfigEvent)) {
```
kaushalkapasi, line 63: will wrapping this `scope.post` inside the IF statement result in any other test issues (i.e. when other tests are mocking out the response to the `/events/batch` endpoint)

jonathannorris: ohh so this one specifically we were setting up two `LocalTestClient` objects for the `should return an empty object if client is not initialized` test above before, which was causing issues when testing these changes as the client setup in the `beforeEach` would never get used. So it made more sense to setup the two `LocalTestClient` in the two tests here.

kaushalkapasi: Got it, sounds good!

kaushalkapasi (commented):
Overall changes look good to me, however we need to ensure that each tests that are using the new `/events/batch` nock handlers are being satisfied or explicitly marked as `not required` to minimize the warnings / "failed test" output

kaushalkapasi (approved):

PR #453, opened 2024-05-31
feat: add ability to load capabilities from env vars

jonathannorris:

### DIFF
--- a/harness/types/capabilities.test.ts
+++ b/harness/types/capabilities.test.ts
@@ -1,0 +1,34 @@
+import { Sdks } from './sdks'
+
+describe('SDKCapabilities Unit Tests', () => {
+    afterEach(() => {
+        jest.resetModules()
+    })
+
+    it('should load capabilities from env variables', () => {
+        process.env.SDK_CAPABILITIES = 'edgeDB,cloud,clientCustomData,sse'
+        process.env.SDKS_TO_TEST = 'NodeJS,Python'
+        const { SDKCapabilities } = jest.requireActual('../capabilities')
+        expect(SDKCapabilities).toEqual({
+            NodeJS: ['edgeDB', 'cloud', 'clientCustomData', 'sse'],
+            Python: ['edgeDB', 'cloud', 'clientCustomData', 'sse'],
+        })
+    })
+
+    it('should load capabilities from env variables as JSON', () => {
+        process.env.SDK_CAPABILITIES =
+            '["edgeDB","cloud","clientCustomData","sse"]'
+        process.env.SDKS_TO_TEST = '["NodeJS","Python"]'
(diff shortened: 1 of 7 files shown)


JamieSinn:
Don't you want to load them from the `/spec` endpoint? https://github.com/DevCycleHQ/test-harness/blob/main/proxies/go/app.go#L34

harness/types/capabilities.test.ts
```diff
@@ -0,0 +1,34 @@
...
+        const { SDKCapabilities } = jest.requireActual('../capabilities')
+        expect(SDKCapabilities).toEqual({
+            NodeJS: ['edgeDB', 'cloud', 'clientCustomData', 'sse'],
+            Python: ['edgeDB', 'cloud', 'clientCustomData', 'sse'],
+        })
+    })
+
+    it('should load capabilities from env variables as JSON', () => {
+        process.env.SDK_CAPABILITIES =
+            '["edgeDB","cloud","clientCustomData","sse"]'
```
ajwootto, line 20: why bother supporting both formats?

harness/types/capabilities.test.ts
```diff
@@ -0,0 +1,34 @@
...
+        process.env.SDKS_TO_TEST = '["NodeJS","Python"]'
+        const { SDKCapabilities } = jest.requireActual('../capabilities')
+        expect(SDKCapabilities).toEqual({
+            NodeJS: ['edgeDB', 'cloud', 'clientCustomData', 'sse'],
+            Python: ['edgeDB', 'cloud', 'clientCustomData', 'sse'],
+        })
+    })
+
+    it('should use default capabilities if env variables are not set', () => {
+        process.env.SDK_CAPABILITIES = 'edgeDB,cloud,clientCustomData,sse'
```
ajwootto, line 30: shouldnt this not be set then?

harness/types/capabilities.test.ts
jonathannorris: basically copied this / wanted to match the `getSDKs()` method: 
image

harness/types/capabilities.test.ts
jonathannorris: The idea is for this to only be used when `SDKS_TO_TEST` is set. The idea is for the various SDKs to be able to set their capabilities when making breaking changes, but didn't want to turn both the existing `SDKS_TO_TEST` + `SDK_CAPABILITIES` into a single new JSON map you need to pass in. Seeing that we are only really going to use this for a single repo testing a new capability I thought this trade-off was fine.

jonathannorris: so this test is testing that basically `SDK_CAPABILITIES` isn't used when `SDKS_TO_TEST` isn't set.

ajwootto (approved):

PR #461, opened 2024-06-07
SSE Constructor Changes for Go SDK

JamieSinn:
SSE Disabled - but the event handler code changed for init events

## Summary by Sourcery

This pull request updates the Go SDK to handle initialization events using a channel for ClientEvent. It also includes dependency upgrades in the go.mod file.

* **Enhancements**:
    - Updated the event handler code to use a channel for ClientEvent instead of a boolean for initialization events.
* **Build**:
    - Upgraded dependencies in go.mod, including github.com/devcyclehq/go-server-sdk/v2 to v2.15.1 and several others.


JamieSinn:
https://github.com/DevCycleHQ/go-server-sdk/pull/234

proxies/go/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/devcyclehq/test-harness/proxies/go
 go 1.22.2
 
 require (
-	github.com/devcyclehq/go-server-sdk/v2 v2.14.0
+	github.com/devcyclehq/go-server-sdk/v2 v2.15.1-0.20240607170116-8bd7750c7b99
```
jonathannorris, line 6: is this a test version?

proxies/go/handler_client.go
```diff
@@ -58,9 +59,9 @@ func clientHandler(w http.ResponseWriter, r *http.Request) {
 		ConfigPollingIntervalMS: time.Duration(reqBody.Options.ConfigPollingIntervalMS * 1000000),
 		EventFlushIntervalMS:    time.Duration(reqBody.Options.EventFlushIntervalMS * 1000000),
 		EnableCloudBucketing:    reqBody.EnableCloudBucketing,
-		OnInitializedChannel:    onInitializedChannel,
+		ClientEventHandler:      onInitializedChannel,
```
jonathannorris, line 62: should the `onInitializedChannel` name change?

JamieSinn: in this case we don't actually care about the name since we don't consume events from it - but it could be a good idea.

jonathannorris (approved):

PR #489, opened 2024-12-13
Count of error responses is not consistent based on SDK speed.

JamieSinn:


harness/features/initialize.local.test.ts
```diff
@@ -105,7 +103,9 @@ describe('Initialize Tests - Local', () => {
             0,
         )
         expect((await variable.json()).data.value).toEqual(0)
-        await wait(1100)
+        scope.get(configRequestUrl).reply(200, testClient.getValidConfig)
```
JamieSinn, line 106: this apparently clears all previous and sets the new return scope. TIL

harness/features/initialize.local.test.ts
```diff
@@ -88,15 +88,13 @@ describe('Initialize Tests - Local', () => {
         const testClient = new LocalTestClient(sdkName)
         const configRequestUrl = testClient.getValidConfigPath()
 
-        scope.get(configRequestUrl).times(2).reply(500)
+        // a 500 is a recoverable error so it will retry the connection and potentially make more requests than expected.
```
suthar26, line 91: why would it make more requests than the polling interval? or is the interval timer only started at successful call?

JamieSinn: Correct - the interval timer starts at the successful call.

suthar26: for all sdks? if that was the case would we have caught this or saw inconsistencies earlier?

suthar26: not fully confident that this wouldnt break other sdks.. but i guess if the tests pass then we can merge it..

JamieSinn:
event flake - should be fine

PR #505, opened 2025-03-14
chore: Use latest stable ruby docker image

JamieSinn:


proxies/ruby/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:3.1 as builder
+FROM ruby as builder
```
kaushalkapasi, line 1: does this also pick latest stable?

kaushalkapasi (approved):

proxies/ruby/Dockerfile
JamieSinn: Yep - same rule as any other docker image

PR #510, opened 2025-03-25
chore: update node.js docker file deps for js-sdks

jonathannorris:


jsalaber (approved):

JamieSinn (approved):
Why so many deps? Would it not make sense to use a more featured docker image base instead?

JamieSinn (approved):

PR #517, opened 2025-05-01
feat: add variablesFeatureId capability and update tests to handle for it

jonathannorris:

### DIFF
--- a/harness/features/allVariables.cloud.test.ts
+++ b/harness/features/allVariables.cloud.test.ts
@@ -6,7 +6,8 @@ import {
     getSDKScope,
 } from '../helpers'
 import { Capabilities } from '../types'
-import { variables } from '../mockData'
+import { getMockedVariables } from '../mockData'
+import { hasCapability } from '../helpers'
 
 describe('allVariables Tests - Cloud', () => {
     const { sdkName, scope } = getSDKScope()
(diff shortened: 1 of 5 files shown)


harness/features/allVariables.cloud.test.ts
```diff
@@ -61,15 +62,24 @@ describe('allVariables Tests - Cloud', () => {
         it('should return a variable map', async () => {
             scope
                 .post(`/client/${client.clientId}/v1/variables`)
-                .reply(200, variables)
+                .reply(
+                    200,
+                    getMockedVariables(
+                        hasCapability(sdkName, Capabilities.variablesFeatureId),
```
JamieSinn, line 68: Why not call this what it actually is - reduced event payload size?

JamieSinn (approved):

harness/features/allVariables.cloud.test.ts
jonathannorris: that's not what this is testing, its testing that `_feature` param is added to the `variables` array for the bucketed config.

JamieSinn: Eh I guess. Just a weird name

JamieSinn (approved):

PR #524, opened 2025-05-15
chore: try 8 core runner

jonathannorris:


jonathannorris:
Seems to basically run at the same speed as the 16 core runner

JamieSinn (approved):

PR #529, opened 2025-06-02
feat: Eval Metadata Feature

JamieSinn:
Templating this here for more bucketing checks later - but this should fix the failures on the Go SDK atm.

### DIFF
--- a/harness/helpers/events.ts
+++ b/harness/helpers/events.ts
@@ -88,6 +88,10 @@ export const expectAggregateEvaluationEvent = ({
         metadata.configLastModified = lastModified
         metadata.clientUUID = expect.any(String)
     }
+    if (hasCapability(sdkName, Capabilities.evalReasons)) {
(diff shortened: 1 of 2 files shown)


harness/helpers/events.ts
```diff
@@ -88,6 +88,10 @@ export const expectAggregateEvaluationEvent = ({
         metadata.configLastModified = lastModified
         metadata.clientUUID = expect.any(String)
     }
+    if (hasCapability(sdkName, Capabilities.evalReasons)) {
+        // Eval reasoning is specific to each test and is outside the scope of testing in general for these events.
+        metadata.eval = expect.any(Object)
```
kaushalkapasi, line 93: should the tests check for:
```suggestion
        metadata.evalReason = expect.any(Object)
```

JamieSinn: Eval is the spec'd parent object you have

PR #534, opened 2025-07-15
chore: update variable tests to check eval object

phaym:
when capabilities are enabled - check for the eval object in the variable return type
- update types package to include variable enums
- keep disabled for now by default, sdks will enable capabilities as they are merged with eval reason updates.

### DIFF
--- a/harness/features/variable.local.test.ts
+++ b/harness/features/variable.local.test.ts
@@ -513,7 +513,7 @@ describe('Variable Tests - Local', () => {
                     expectAggregateDefaultEvent({
                         body: eventResult.body,
                         variableKey: key,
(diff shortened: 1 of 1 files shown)


harness/features/variable.local.test.ts
```diff
@@ -192,6 +206,7 @@ describe('Variable Tests - Local', () => {
                         wrongTypeDefault === '1'
                             ? VariableType.string
                             : VariableType.number,
+                        DEFAULT_REASON_DETAILS.USER_NOT_TARGETED,
```
suthar26, line 209: does this need to be checked for capability?

phaym: the expectDefaultValue function handles that

PR #535, opened 2025-07-15
chore: exclude of-node from allVariables/allFeatures tests

phaym:
adding allVariables and allFeatures capabilities so of-node tests can skip these test files
- these functions are not supported by OF and the proxy was just calling devcycle sdk functions, which conflicted with capability for evalReason and caused tests to fail
- lets test-harness pass for of-node, in future it will have its own evalReasons capability for variable call only and i can undo this change

### DIFF
--- a/harness/features/allFeatures.local.test.ts
+++ b/harness/features/allFeatures.local.test.ts
@@ -10,6 +10,7 @@ import { expectedFeaturesVariationOn } from '../mockData'
 describe('allFeatures Tests - Local', () => {
     const { sdkName, scope } = getSDKScope()
 
+    //TODO: when tests are updated for OF-NodeJS eval capability, remove this allFeatures capability wrapper
     describeCapability(sdkName, Capabilities.allFeatures)(sdkName, () => {
         describe('uninitialized client', () => {
             let testClient: LocalTestClient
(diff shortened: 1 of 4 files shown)


harness/features/allFeatures.local.test.ts
```diff
@@ -1,10 +1,12 @@
-import { LocalTestClient, getSDKScope, hasCapability } from '../helpers'
+import { LocalTestClient, getSDKScope, hasCapability, describeCapability } from '../helpers'
 import { Capabilities } from '../types'
 import { expectedFeaturesVariationOn } from '../mockData'
 
 describe('allFeatures Tests - Local', () => {
     const { sdkName, scope } = getSDKScope()
 
+    describeCapability(sdkName, Capabilities.allFeatures)(sdkName, () => {
```
jonathannorris, line 8: All the code within this should be indented.

harness/features/allVariables.local.test.ts
```diff
@@ -3,13 +3,16 @@ import {
...
     hasCapability,
+    describeCapability,
 } from '../helpers'
 import { Capabilities } from '../types'
 import { getMockedVariables } from '../mockData'
 
 describe('allVariables Tests - Local', () => {
     const { sdkName, scope } = getSDKScope()
 
+    describeCapability(sdkName, Capabilities.allVariables)(sdkName, () => {
```
jonathannorris, line 14: Same here, all the containing code should be indented.

phaym: yep its done, ai generated and I didnt save to format

phaym:
we still want to run this test for OF so i'll come up with a different solution

jonathannorris (approved):

PR #537, opened 2025-07-17
chore: tests for node openfeature eval reasons

phaym:
update test-harness for openfeature node eval reason tests
- return the reason from the OF-node proxy
- handle OF case for expectations since the sdk only returns a single 'reason' field, as opposed to the object returned by our sdk
- enable evalReason for openfeature - there was already logic to determine DEFAULT vs TARGETING_MATCH. Since I updated the proxy this value is now being returned even from existing release. Once this is released, it will be coming from wasm instead with more possible reasons

### DIFF
--- a/harness/features/variable.cloud.test.ts
+++ b/harness/features/variable.cloud.test.ts
@@ -367,9 +367,9 @@ describe('Variable Tests - Cloud', () => {
                 `should return defaulted ${type} %s if mock server returns an internal error, \
                 after retrying 5 times`,
                 async (method) => {
-                    const hasEvalReason = hasCapability(
+                    const hasCloudEvalReason = hasCapability(
                         sdkName,
-                        Capabilities.evalReason,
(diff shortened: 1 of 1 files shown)


harness/features/variable.cloud.test.ts
```diff
@@ -50,6 +51,16 @@ describe('Variable Tests - Cloud', () => {
...
         )
     }
+    function getEvalReason(
+        sdkName: string,
+        reason: string,
+        details?: string,
+        target_id?: string,
+    ) {
+        return sdkName === 'OF-NodeJS'
+            ? { reason }
```
jonathannorris, line 61: we are looking to potentially add support for the other properties in Android OF Provider, but not a required thing for this PR: https://github.com/DevCycleHQ/android-client-sdk/pull/227#discussion_r2216603453

jonathannorris (approved):

PR #553, opened 2025-07-30
chore: update python capabilities for eval reasons

phaym:
python released with eval reason - add capabilities to be run by default

### DIFF
--- a/harness/types/capabilities.ts
+++ b/harness/types/capabilities.ts
@@ -61,6 +61,9 @@ let sdkCapabilities: { [key: string]: string[] } = {
         Capabilities.edgeDB,
         Capabilities.clientCustomData,
         Capabilities.v2Config,
+        Capabilities.cloudEvalReason,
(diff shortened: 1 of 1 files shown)


harness/types/capabilities.ts
```diff
@@ -61,6 +61,9 @@ let sdkCapabilities: { [key: string]: string[] } = {
         Capabilities.edgeDB,
         Capabilities.clientCustomData,
         Capabilities.v2Config,
+        Capabilities.cloudEvalReason,
+        Capabilities.evalReason,
+        Capabilities.eventsEvalReason,
```
kaushalkapasi, line 66: also `flagMetadata`?

phaym: no still not yet because OF not implemented

kaushalkapasi: ah ok

kaushalkapasi (approved):

